PR #11, opened 2020-07-22
support non project root language servers

UziTech:
moved from https://github.com/atom/atom-languageclient/pull/250

see issue https://github.com/atom/atom-languageclient/issues/233

### DIFF
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -234,6 +234,17 @@ export default class AutoLanguageClient {
     return configuration;
   }
 
+  // Determine the project path
+  protected async determineProjectPath(editor: TextEditor): Promise<string | null> {
+    const filePath = editor.getPath();
+    if (filePath == null) {
+      return null;
+    }
+    return atom.project.getDirectories()
+      .map((d) => Utils.normalizePath(d.getPath()))
+      .find((d) => filePath.startsWith(d)) || null;
+  }
+
   // Helper methods that are useful for implementors
   // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


aminya:
This branch is removed. Could you revive it?

aminya:
@UziTech Could you rebase this? This is needed for flow-ide

https://github.com/steelbrain/flow-ide

lib/server-manager.ts
```diff
@@ -258,27 +257,10 @@ export class ServerManager {
...
-  }
-
-  public updateNormalizedProjectPaths(): void {
-    this._normalizedProjectPaths = atom.project.getDirectories().map((d) => this.normalizePath(d.getPath()));
-  }
-
-  public normalizePath(projectPath: string): string {
-    return !projectPath.endsWith(path.sep) ? path.join(projectPath, path.sep) : projectPath;
-  }
-
```
aminya, line 276: I think we should just move this code to auto-languageclient so the user can override them. Maybe there is a better way to make this configurable instead of moving the methods

lib/auto-languageclient.ts
```diff
@@ -263,6 +274,7 @@ export default class AutoLanguageClient {
       (filepath) => this.filterChangeWatchedFiles(filepath),
       this.reportBusyWhile,
       this.getServerName(),
+      (e) => this.determineProjectPath(e),
```
aminya, line 277: Instead of all these, we can probably provide a method that is passed down to `this._serverManager`.

Passed as an argument:
```js
      this.customDetermineProject
```

Later we will check if this method is defined to use that instead.

UziTech:
I don't have this branch anymore. It would be easier for you to create a new PR that does this.

aminya:
Yeah, I think it's easier to start over

PR #88, opened 2020-10-19
fix: rename to atom-languageclient

UziTech:


UziTech:
I'll create another PR rebased with the squashed commits.

PR #89, opened 2020-10-19
fix: rename to atom-languageclient

UziTech:


UziTech:
after this is merged we should be ready to release v1.0.0

Do we want to to add anything to the readme about the change?

aminya:
> after this is merged we should be ready to release v1.0.0
> 
> Do we want to to add anything to the readme about the change?

I think just saying that the repository is moved to this organization.

Also, make sure to update the links in package.json.

PR #93, opened 2020-10-20
chore: add bump_deps action using ncu

aminya:
This uses ncu instead of the dependent bot which offers more control and flexibility.

We need to remove the dependent bot after this.

Closes #92 Closes #91


UziTech:
This action won't run the tests when the PR is created if we use `GITHUB_TOKEN` to create it.

aminya:
I'll merge this since we have been using it in other repositories. Testing it in action requires waiting until the bot getting actually triggered.

aminya:
> This action won't run the tests when the PR is created if we use `GITHUB_TOKEN` to create it.

Should I remove GITHUB_TOKEN?

UziTech:
We should use a different token.

UziTech:
it defaults to `GITHUB_TOKEN` if we don't provide one.

aminya:
I am OK with using a token. These actions are safe and are widely used by the community.

UziTech:
If you have access to @atom-ide-community-bot we should use that account

UziTech:
You could create an organization secret with the PAT something like `GH_TOKEN`

aminya:
Unfortunately, I don't have access to that bot. I can create the token by my own account.

UziTech:
that will work too.

UziTech:
FYI here is the information about why it doesn't run tests with `GITHUB_TOKEN`

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token

aminya:
The action suggests creating a GitHub app: https://github.com/peter-evans/create-pull-request/blob/master/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens

UziTech:
That could work too.

aminya:
The lock file on the master seems old. I'll remove and generate it again.

PR #98, opened 2020-10-29
fix: use native fuzzaldrin-plus-fast

aminya:
This replaces fuzzaldrin-plus with fuzzaldrin-plus-fast which is our fast native C++ module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   },
   "atomTestRunner": "./test/runner",
   "dependencies": {
-    "fuzzaldrin-plus": "^0.6.0",
+    "fuzzaldrin-plus-fast": "^1.0.1",
     "rimraf": "^3.0.2",
     "vscode-jsonrpc": "5.0.1",
     "vscode-languageserver-protocol": "3.15.3",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -26,7 +26,6 @@
     "@atom/mocha-test-runner": "^1.6.0",
     "@types/atom": "^1.40.5",
     "@types/chai": "^4.2.14",
-    "@types/fuzzaldrin-plus": "0.6.1",
```
UziTech, line 29: Why remove the types? fuzzaldrin-plus-fast should support the exact same types.

aminya: Because the native package has its own types. We don't need a separate installation anymore.

UziTech (approved):

aminya:
@UziTech before you merge this, I want to include prebuilds for electroin@7 just in case.

UziTech:
ok, and I was going to add semantic-release. Is there any other PRs we want to get into the next release?

aminya:
> include prebuilds for electron@7 just in case

So, as I was correct, the prebuild of electron 7 is the same as electron 6. My prebuildify does not generate anything new for electron 7. It is truly forward compatible.

UziTech:
Should this be a patch release? I wouldn't consider this a new feature since it is just replacing something.

aminya:
Yes, it is a fix release.

PR #101, opened 2020-10-30
Fixes textEdit completions

### LINKED ISSUE #18: Fix text completions
We should start with this: https://github.com/atom/atom-languageclient/pull/254

UziTech:
rebase of https://github.com/atom/atom-languageclient/pull/254

fixes #18


UziTech:
This should have tests to make sure it does what it is supposed to before it gets merged.

aminya (commented):
Would be nice to get this in too

UziTech:
I'm not sure this is needed. it seems like this fixes the same thing as #100

PR #106, opened 2020-12-12
fix: set workspaceFolders to null when there is no workspace

aminya:
Based on  https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize
```ts

	/**
	 * The workspace folders configured in the client when the server starts.
	 * This property is only available if the client supports workspace folders.
	 * It can be `null` if the client supports workspace folders but none are
	 * configured.
	 *
	 * @since 3.6.0
	 */
	workspaceFolders?: WorkspaceFolder[] | null;
```

https://github.com/vuejs/vetur/issues/2561#issuecomment-743719403


UziTech (approved):
LGTM

PR #117, opened 2021-01-17
feat: API for registering custom callbacks for Commands

deankevorkian:
Followup to #12 , discussions made there and [there](https://github.com/atom/atom-languageclient/issues/183).

A developer can register a callback that happens when an `executeCommand` is fired for `DoBlaBlaJavaThingyTypeBeat`. If a command was not registered with a callback by the developer - let that `Command` reach the server. Otherwise - fire the callback that was registered by the developer.

This brings package developers full control as to how `Command`s are executed "globally".
i.e - they can register a command and a callback for it, and have parts of their package (like code actions, or other custom areas) call those `Command`s and have them executed just like VSCode does today. But, this will probably be mainly needed/used for code actions, [where the standard isn't explicit with `Command`s are executed](https://github.com/microsoft/language-server-protocol/issues/430).

I could also use some help testing this with ide-java as I'm not used to building and debugging extensions. I can do this myself but if anyone can easily test this let me know.

@UziTech

### DIFF
--- a/lib/adapters/command-execution-adapter.ts
+++ b/lib/adapters/command-execution-adapter.ts
@@ -4,7 +4,7 @@ import { LanguageClientConnection } from "../main";
 export type CommandCustomCallbackFunction = (command: ExecuteCommandParams) => Promise<any | void>;
 
 export default class CommandExecutionAdapter {
-    private static commandsCustomCallbacks: Map<string, CommandCustomCallbackFunction> = new Map<string, CommandCustomCallbackFunction>();
+    private static commandsCustomCallbacks = new Map<string, CommandCustomCallbackFunction>();
 
     public static canAdapt(serverCapabilities: ServerCapabilities): boolean {
       return serverCapabilities.executeCommandProvider != null;
@@ -14,14 +14,14 @@ export default class CommandExecutionAdapter {
         this.commandsCustomCallbacks.set(command, callback);
     }
 
-    public static async executeCommand(connection: LanguageClientConnection, command: string, commandArgs?: any[] | undefined): Promise<any | void> {
+    public static async executeCommand(connection: LanguageClientConnection, command: string, commandArgs?: any[]): Promise<any | void> {
         const executeCommandParams = CommandExecutionAdapter.createExecuteCommandParams(command, commandArgs);
         const commandCustomCallback = this.commandsCustomCallbacks.get(command);
 
         return commandCustomCallback != null ? await commandCustomCallback(executeCommandParams) : await connection.executeCommand(executeCommandParams);
     }
 
-    private static createExecuteCommandParams(command: string, commandArgs?: any[] | undefined): ExecuteCommandParams {
+    private static createExecuteCommandParams(command: string, commandArgs?: any[]): ExecuteCommandParams {
         return {
             command: command,
             arguments: commandArgs
--- a/lib/adapters/command-execution-adapter.ts
+++ b/lib/adapters/command-execution-adapter.ts
@@ -18,7 +18,7 @@ export default class CommandExecutionAdapter {
         const executeCommandParams = CommandExecutionAdapter.createExecuteCommandParams(command, commandArgs);
         const commandCustomCallback = this.commandsCustomCallbacks.get(command);
 
-        return commandCustomCallback != null ? await commandCustomCallback(executeCommandParams) : await connection.executeCommand(executeCommandParams);
+        return commandCustomCallback !== undefined ? await commandCustomCallback(executeCommandParams) : await connection.executeCommand(executeCommandParams);
     }
 
     private static createExecuteCommandParams(command: string, commandArgs?: any[]): ExecuteCommandParams {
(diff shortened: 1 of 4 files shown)


lib/adapters/command-execution-adapter.ts
```diff
@@ -0,0 +1,26 @@
+import { ExecuteCommandParams } from "../languageclient";
+import { LanguageClientConnection } from "../main";
+
+export type CommandCustomCallbackFunction = (command: ExecuteCommandParams) => Promise<any | void>;
+
+export default class CommandExecutionAdapter {
+    private static commandsCustomCallbacks: Map<string, CommandCustomCallbackFunction> = new Map<string, CommandCustomCallbackFunction>();
+
+    public static registerCustomCallbackForCommand(command: string, callback: CommandCustomCallbackFunction): void {
```
UziTech, line 9: How are we exposing this function to the package author?

deankevorkian: Come to think about it, I just assumed one would invoke that method directly via the `CommandExecutionAdapter` static class, but I guess we could also have a method in `AutoLanguageClient` that simply calls that adapter. What'd you had in mind?

lib/adapters/command-execution-adapter.ts
```diff
@@ -0,0 +1,30 @@
+import { ExecuteCommandParams, ServerCapabilities } from "../languageclient";
+import { LanguageClientConnection } from "../main";
+
+export type CommandCustomCallbackFunction = (command: ExecuteCommandParams) => Promise<any | void>;
+
+export default class CommandExecutionAdapter {
+    private static commandsCustomCallbacks: Map<string, CommandCustomCallbackFunction> = new Map<string, CommandCustomCallbackFunction>();
+
+    public static canAdapt(serverCapabilities: ServerCapabilities): boolean {
+      return serverCapabilities.executeCommandProvider != null;
```
UziTech, line 10: if `serverCapabilities.executeCommandProvider` is `false` this would return `true`. Shouldn't we just check for `true`?
```suggestion
      return serverCapabilities.executeCommandProvider === true;
```

deankevorkian: `executeCommandProvider` is of type `ExecuteCommandOptions`unlike other providers that are either boolean or an object of options (depending on protocol version). I think the check is the only way of knowing if it was supplied or not.

UziTech: I see

UziTech (approved):

lib/adapters/command-execution-adapter.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public static registerCustomCallbackForCommand(command: string, callback: CommandCustomCallbackFunction): void {
+        this.commandsCustomCallbacks.set(command, callback);
+    }
+
+    public static async executeCommand(connection: LanguageClientConnection, command: string, commandArgs?: any[] | undefined): Promise<any | void> {
```
aminya, line 17:
```suggestion
    public static async executeCommand(connection: LanguageClientConnection, command: string, commandArgs?: any[]): Promise<any | void> {
```

lib/adapters/command-execution-adapter.ts
```diff
@@ -0,0 +1,30 @@
...
+        const executeCommandParams = CommandExecutionAdapter.createExecuteCommandParams(command, commandArgs);
+        const commandCustomCallback = this.commandsCustomCallbacks.get(command);
+
+        return commandCustomCallback != null ? await commandCustomCallback(executeCommandParams) : await connection.executeCommand(executeCommandParams);
```
aminya, line 21:
```suggestion
        return commandCustomCallback !== undefined ? await commandCustomCallback(executeCommandParams) : await connection.executeCommand(executeCommandParams);
```

lib/adapters/command-execution-adapter.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    private static createExecuteCommandParams(command: string, commandArgs?: any[] | undefined): ExecuteCommandParams {
```
aminya, line 24:
```suggestion
    private static createExecuteCommandParams(command: string, commandArgs?: any[]): ExecuteCommandParams {
```

lib/adapters/command-execution-adapter.ts
```diff
@@ -0,0 +1,30 @@
+import { ExecuteCommandParams, ServerCapabilities } from "../languageclient";
+import { LanguageClientConnection } from "../main";
+
+export type CommandCustomCallbackFunction = (command: ExecuteCommandParams) => Promise<any | void>;
+
+export default class CommandExecutionAdapter {
+    private static commandsCustomCallbacks: Map<string, CommandCustomCallbackFunction> = new Map<string, CommandCustomCallbackFunction>();
```
aminya, line 7:
```suggestion
    private static commandsCustomCallbacks = new Map<string, CommandCustomCallbackFunction>();
```

aminya (approved):
We should update our Readme too, but I am OK with deferring it to another PR since our documentation needs a revamp.

PR #118, opened 2021-01-27
chore: use pnpm + misc chore 

aminya:


.gitattributes
```diff
@@ -0,0 +1,6 @@
+* text=auto
+
+# don't diff machine generated files
+dist/ -diff
```
UziTech, line 4: Is this supposed to be the `build` folder?
```suggestion
build/ -diff
```

Issue #115, opened 2021-01-07
Set up Prettier


illright:
As I've been working with the codebase, I've noticed how most lines are around 100 characters long, which makes the codebase rather difficult to navigate (makes me collapse the Tree View at times just so everything could fit on my screen). That is an issue on its own, but I think we could resolve it and even more by installing Prettier for this project.

A bit of insight in case you're unfamiliar with Prettier: it's a universal code formatter that is opinionated and unconfigurable by design. It helps to keep line lengths at bay as well as resolve endless debates about code formatting. I admit, there are a couple of formatting decisions in Prettier that I dislike, but when installing it for my projects, I've come to realize that this is a sacrifice well worth settling code formatting once and for all.

Not a priority, just as a future consideration. What do you think?

aminya:
There is a prettier config that the organization uses for all of the packages. I will make a PR and will add it here. 

https://github.com/atom-community/atom-ide-template/blob/master/prettier.config.js

illright:
120 characters on the print width though? That's rather extreme. In my projects I've found 100 to be the comfortable deadline for line lengths, but print width is supposed to be a guideline, not a strict deadline, and having that at 120 means the lines will get very long. Could we alter that to 80-90?

UziTech:
I prefer long lines to breaking things up into multiple lines

```js
const someLongString =
  "Some very very very very very very very very very very very long string"
```

That is a lot harder for me to skim and understand what is happening than having it on the same line.

aminya:
120 is a good balance inside Atom. In a normal zoom inside Atom, if you have Minimap, GitHub pane, and TreeView pane open, you can still see 120 characters.

illright:
The problem is, everyone has different screens :) I'm on a 14" laptop, so when I have just the Tree View open, the last column that fits on my screen is 85. That's why, as I was working with the codebase, I felt really lost and had to hide the tree view to just have the editor fullscreen and only then was I able to properly read the code.

I understand what you're saying about function signature verbosity, I did find myself breaking up lines frequently with TypeScript, but I personally think it pays off.

Feels like I am outnumbered here :p. Could we find a compromise?

UziTech:
I don't think one way is **better** than the other so I think we should go for consistency with other repos in the organization.

The only style component I have a strong opinion about is using tabs instead of spaces mainly for file size, and developer accessibility. Why would you demand a specific indent size when you can easily allow every developer to specify their own preferable indent size? (but I digress).

[...]

PR #121, opened 2021-02-07
fix: use types from atom-ide-base

aminya:
### Description
- This PR uses the types from atom-ide-base.
- This fixes the typescript errors when someone uses atom-languageclient

I compared the types in atom-languageclient with the types we had here in atom-ide-base, and I added the missing ones into atom-ide-base. 
See https://github.com/atom-community/atom-ide-base/pull/57

We no longer need to maintain the types in two places.

### DIFF
--- a/lib/adapters/notifications-adapter.ts
+++ b/lib/adapters/notifications-adapter.ts
@@ -10,9 +10,10 @@ import {
   NotificationOptions,
   NotificationExt,
 } from 'atom';
-import {
-  NotificationButton,
-} from 'atom-ide';
+
+export interface NotificationButton {
+  text: string
+}
 
 /** Public: Adapts Atom's user notifications to those of the language server protocol. */
 export default class NotificationsAdapter {
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -3,6 +3,7 @@ import * as ls from './languageclient';
 import * as rpc from 'vscode-jsonrpc';
 import * as path from 'path';
 import * as atomIde from 'atom-ide';
+import type { OutlineProvider } from 'atom-ide-base'
 import * as linter from 'atom/linter';
 import Convert from './convert.js';
 import ApplyEditAdapter from './adapters/apply-edit-adapter';
@@ -539,7 +540,7 @@ export default class AutoLanguageClient {
   }
 
   // Outline View via LS documentSymbol---------------------------------
-  public provideOutlines(): atomIde.OutlineProvider {
+  public provideOutlines(): OutlineProvider {
     return {
       name: this.name,
       grammarScopes: this.getGrammarScopes(),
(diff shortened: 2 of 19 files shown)


aminya:
The error here is an internal error of atom-ide-base because JSX is defined by React (or Etch). Here React is not used, so the code throws an error. I think the solution is to use `React.JSX.Element` in the types. 

https://github.com/atom-community/atom-ide-base/blob/8e952b465a6485eee76881cc201108e5bd853668/types-packages/datatip.d.ts#L72

lib/auto-languageclient.ts
```diff
@@ -513,7 +514,7 @@ export default class AutoLanguageClient {
   protected onDidInsertSuggestion(_arg: ac.SuggestionInsertedEvent): void { }
 
   // Definitions via LS documentHighlight and gotoDefinition------------
-  public provideDefinitions(): atomIde.DefinitionProvider {
+  public provideDefinitions(): DefinitionProvider {
```
aminya, line 517: `wordRegExp` is missing here in atom-languageclient

test/adapters/datatip-adapter.test.ts
```diff
@@ -50,14 +50,17 @@ describe('DatatipAdapter', () => {
...
-        invariant(snippet.type === 'snippet');
-        expect((snippet as any).grammar.scopeName).equal('text.plain.null-grammar');
-        expect(snippet.value).equal('test snippet');
+          const snippet = datatip.markedStrings[1];
+          expect(snippet.type).equal('snippet');
+          invariant(snippet.type === 'snippet');
+          expect((snippet as any).grammar.scopeName).equal('text.plain.null-grammar');
+          expect(snippet.value).equal('test snippet');
+        }
+        // else the React variant
```
aminya, line 63: atom-languageclient only considered the markedString version as Datatip type

lib/adapters/notifications-adapter.ts
```diff
@@ -10,9 +10,10 @@ import {
   NotificationOptions,
   NotificationExt,
 } from 'atom';
-import {
-  NotificationButton,
-} from 'atom-ide';
+
+export interface NotificationButton {
```
aminya, line 14: NotificationButton is not an ide type so I moved it here

lib/auto-languageclient.ts
```diff
@@ -2,7 +2,7 @@ import * as cp from 'child_process';
 import * as ls from './languageclient';
 import * as rpc from 'vscode-jsonrpc';
 import * as path from 'path';
-import * as atomIde from 'atom-ide';
+import type { OutlineProvider, Outline, DefinitionProvider, DefinitionQueryResult, FindReferencesProvider, FindReferencesReturn, Datatip, DatatipService, TextEdit, RangeCodeFormatProvider, FileCodeFormatProvider, OnSaveCodeFormatProvider, OnTypeCodeFormatProvider, CodeAction, CodeHighlightProvider, CodeActionProvider, Diagnostic, RefactorProvider, BusySignalService, SignatureHelpRegistry, ConsoleService } from 'atom-ide-base';
```
aminya, line 5: #115 Will take care of formatting later

UziTech (approved):

Issue #110, opened 2020-12-20
Update jsonrpc and languageserver dependencies 


aminya:
New updates have come for vscode dependencies. We need to apply the upgrades. The new languageserver-protocol has new features that we should add to atom-languageclient.

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#version_3_16_0
https://github.com/microsoft/vscode-languageserver-node#3160-protocol-600-json-rpc-700-client-and-700-server
```
 vscode-jsonrpc                   5.0.1  →   6.0.0
 vscode-languageserver-protocol  3.15.3  →  3.16.0
 vscode-languageserver-types     3.15.1  →  3.16.0
```

PR #112, opened 2021-01-02
Upgrade to LSP 3.16

illright:
I'm currently working on upgrading the LSP version. For now I have just set up some foundations, fixed compiler errors from the update, and now I'm gonna start implementing the new features. Figured it'd be good to get feedback as I'm working since this is my first dive into the Atom ecosystem

Closes #110

PR #122, opened 2021-02-07
feat: Upgrade to LSP 3.16 

aminya:
Fixes https://github.com/atom-community/atom-languageclient/issues/110

The first commit of #112 is cherry-picked

I added tests for the new autocomplete feature added.

### DIFF
--- a/lib/adapters/autocomplete-adapter.ts
+++ b/lib/adapters/autocomplete-adapter.ts
@@ -25,13 +25,10 @@ import { Suggestion, TextSuggestion, SnippetSuggestion } from "../types/autocomp
  * Defines the behavior of suggestion acceptance.
  * Assume you have "cons|ole" in the editor (`|` is the cursor position)
  * and the autocomplete suggestion is `const`.
- * - insert mode: const|ole
- * - replace mode: const|
+ * - if `false` -> the edits are inserted : const|ole
+ * - if `true`` -> the edits are replaced: const|
  */
-enum InsertMode {
-  insert = 'insert',
-  replace = 'replace',
-}
+type ShouldReplace = boolean;
 
 /**
  * Holds a list of suggestions generated from the CompletionItem[]
@@ -87,7 +84,7 @@ export default class AutocompleteAdapter {
    * @param onDidConvertCompletionItem An optional function that takes a {CompletionItem},
    *   an {atom$AutocompleteSuggestion} and a {atom$AutocompleteRequest}
    *   allowing you to adjust converted items.
-   * @param insertMode The behavior of suggestion acceptance (see {InsertMode}).
+   * @param shouldReplace The behavior of suggestion acceptance (see {ShouldReplace}).
    * @returns A {Promise} of an {Array} of {atom$AutocompleteSuggestion}s containing the
    *   AutoComplete+ suggestions to display.
    */
@@ -96,7 +93,7 @@ export default class AutocompleteAdapter {
     request: ac.SuggestionsRequestedEvent,
     onDidConvertCompletionItem?: CompletionItemAdjuster,
     minimumWordLength?: number,
-    insertMode: InsertMode = InsertMode.insert
+    shouldReplace: ShouldReplace = false,
   ): Promise<ac.AnySuggestion[]> {
     const triggerChars =
       server.capabilities.completionProvider != null
@@ -117,7 +114,7 @@ export default class AutocompleteAdapter {
       request,
       triggerChar,
       triggerOnly,
-      insertMode,
+      shouldReplace,
       onDidConvertCompletionItem
     )
 
@@ -157,7 +154,7 @@ export default class AutocompleteAdapter {
     request: ac.SuggestionsRequestedEvent,
     triggerChar: string,
     triggerOnly: boolean,
-    insertMode: InsertMode,
+    shouldReplace: ShouldReplace,
     onDidConvertCompletionItem?: CompletionItemAdjuster
   ): Promise<Suggestion[]> {
     const cache = this._suggestionCache.get(server)
@@ -199,7 +196,7 @@ export default class AutocompleteAdapter {
       completions,
       request,
       triggerColumns,
-      insertMode,
+      shouldReplace,
       onDidConvertCompletionItem
     )
     this._suggestionCache.set(server, {
@@ -358,7 +355,7 @@ export default class AutocompleteAdapter {
    * @param completionItems An {Array} of {CompletionItem} objects or a {CompletionList} containing completion
    *   items to be converted.
    * @param request The {atom$AutocompleteRequest} to satisfy.
-   * @param insertMode The behavior of suggestion acceptance (see {InsertMode}).
+   * @param shouldReplace The behavior of suggestion acceptance (see {ShouldReplace}).
    * @param onDidConvertCompletionItem A function that takes a {CompletionItem}, an {atom$AutocompleteSuggestion}
    *   and a {atom$AutocompleteRequest} allowing you to adjust converted items.
    * @returns A {Map} of AutoComplete+ suggestions ordered by the CompletionItems sortText.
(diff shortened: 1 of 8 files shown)


UziTech:
I'll add a few tests to this before approving.

lib/adapters/autocomplete-adapter.ts
```diff
@@ -429,18 +451,29 @@ export default class AutocompleteAdapter {
...
+      range = textEdit.range;
+    } else if (insertMode === InsertMode.insert) {
+      range = textEdit.insert;
+    } else {
+      range = textEdit.replace;
+    }
+
+    if (range.start.character !== triggerColumns[0]) {
+      const atomRange = Convert.lsRangeToAtomRange(range);
+      suggestion.customReplacmentPrefix = editor.getTextInBufferRange([atomRange.start, originalBufferPosition]);
```
aminya, line 476: This code needs a test

lib/adapters/autocomplete-adapter.ts
```diff
@@ -22,6 +24,18 @@ import {
...
+ * Defines the behavior of suggestion acceptance.
+ * Assume you have "cons|ole" in the editor (`|` is the cursor position)
+ * and the autocomplete suggestion is `const`.
+ * - insert mode: const|ole
+ * - replace mode: const|
+ */
+enum InsertMode {
+  insert = 'insert',
+  replace = 'replace',
+}
```
aminya, line 37: An enum with only two modes is a sign that we should just use a boolean instead.

illright: Not really. Neither of these options is more important, and `not insert` does not imply `replace` even in terms of meaning, so I'd argue that for code clarity it'd be best to keep the explicit enum

lib/adapters/autocomplete-adapter.ts
illright: I am, but seeing as I'm a little busy right now, it might take a couple days before I can get to it. If that's fine with you, then sure, I'll add it.

lib/adapters/autocomplete-adapter.ts
UziTech: > `not insert` does not imply `replace`

It actually does according to the code:

```js
if (insertMode === InsertMode.insert) {
  range = textEdit.insert;
} else {
  range = textEdit.replace;
}
```

illright: Yeah, I meant in terms of meaning. Having it a boolean variable doesn't help to understand what "not insert" means without looking at the code

aminya: Using `enum` actually makes the code more complex. If you want clarity, you can add more documentation for the parameter. You can name the boolean value as `shouldReplace`. If it is false, it means to insert the text.

illright: > Using `enum` actually makes the code more complex.

How so, apart from more to type?

aminya: Instead of passing a simple `1-bit Boolean`, you now require using the Enum which is an object of strings stored on the heap.

illright: Well, computational complexity, yes, but we're using JavaScript, a high-level language. It's not that we should write inefficient code, it's just that these kinds of inefficiencies are a drop to the ocean. And it's up to the JS runtime to perform optimizations for this kind of code. So I think that for high-level languages, code clarity is the priority.

But for code clarity, how are booleans better than enums?

P.S. I believe that booleans are not 1-bit, but at least 1 byte to be able to be efficiently addressed and not break alignment.

illright: Please, don't make changes that are being discussed before an agreement is reached. It's not very nice as it implies that the decision was unilaterally made by you, regardless of other opinions.

aminya: I don't understand how enum has more clarity. This is a simple option with only two states. If you see the parameter name, you easily understand that you should pass `true` or `false`.

illright: For starters, JS doesn't support named parameters, so imagine a function call like `getSuggestions(text, true)`. When you see code like that, what do you think this `true` means? Now compare that to `getSuggestions(text, InsertMode.replace)`

aminya: `enum` is not a JavaScript type. It is a TypeScript extension.

illright: True, but what of it?

aminya: This makes `enum` inherently complex. JavaScript developers don't have any notion of `enum`, and so this is just making things complex for an option that could be a simple `boolean`.

Even if there were multiple states, the Union type of TypeScript was the simpler and clearer type to work with.

illright: We're writing code in TypeScript, not leveraging its functionality for the sake of "pure-JS devs" doesn't sound reasonable. And what about the named parameters issue?

aminya: The named parameter is another story (more general than boolean vs enum). 

The solution is to insert documentation whenever you want to make the callsite more verbose.
```js
getSuggestions(/* text */ text, /* insertMode */ true)
```

I have been thinking to make an Atom package for this purpose.

illright: Yeah, but frankly this looks awful to me, and is a huge burden in terms of navigating code. It is done well in JetBrains IDEs that kind of "inject" the name into the code without it actually being there, but that's IDE-specific, and we don't want to write code that is only easy to manage in one particular kind of IDE. 

That's why the call clarity is a very strong argument in favor of the `enum`, in my opinion. Maybe we could get a third-person view at this?

aminya: >  a very strong argument in favor of the enum,

That is not "very strong" to me. I gave you stronger arguments why this is not preferred. 

Regarding the named parameters, we have been already using those:
https://github.com/atom-community/atom-ide-outline/blob/77c5e26519d18d026320cd66bd97b98f8c25c1b8/src/main.ts#L39

> Maybe we could get a third-person view at this?

We have more important things to work on. If we want to hold an election for every single review, nothing will go forward.

illright: Well, your arguments are not strong to me. Seeing as you already changed the code to booleans, fine, you can keep it that way, but I must confess that this was yet another drop of demotivation for me to continue working on this. I don't think I'll be doing the tests for this PR. Feel free to add them.

aminya: What I have learned by contributing to hundreds of projects on GitHub is to follow the coding paradigms/conventions that a specific project is using. This has given my PRs a very high chance of getting approved. No matter if I am an OOP guy or a functional guy, or if I like TypeScript enums more than booleans, I know how to follow what the reviewers want. At the end of the day, it is the "main contribution" that matters the most.

Tony Van Eerd nails this "People already have enough paradigms" greatly.
https://youtu.be/GPP64opjy_Y?t=1668

aminya:
I rebased this incorrectly. I will fix it tomorrow.

test/adapters/autocomplete-adapter.test.ts
```diff
@@ -90,6 +98,21 @@ describe("AutoCompleteAdapter", () => {
       documentation: "should not appear",
       sortText: "zzz",
     }),
+    createCompletionItem("textToBeEdited1", {
```
aminya, line 101: Do these seem like good data to test the added code path?

aminya:
Requires #134

test/adapters/autocomplete-adapter.test.ts
```diff
@@ -622,8 +647,33 @@ describe("AutoCompleteAdapter", () => {
...
+            customRequest,
+            undefined,
+            undefined,
+            false
+          )
+
+          expect(results4[0].displayText).equals("align")
+          expect((results4[0] as TextSuggestion).text).equals("hello world")
+          expect(results4[0].replacementPrefix).equals("")
+          expect((results4[0] as any).customReplacmentPrefix).equals(undefined)
```
aminya, line 676: Not sure if this is expected

test/adapters/autocomplete-adapter.test.ts
```diff
@@ -548,6 +548,31 @@ describe("AutoCompleteAdapter", () => {
...
+            customRequest,
+            undefined,
+            undefined,
+            true
+          )
+
+          expect(results4[0].displayText).equals("align")
+          expect((results4[0] as TextSuggestion).text).equals("hello world")
+          expect(results4[0].replacementPrefix).equals("")
+          expect((results4[0] as any).customReplacmentPrefix).equals(undefined)
```
aminya, line 574: Not sure if this is expected

UziTech (approved):
LGTM

PR #123, opened 2021-02-20
fix: fix type errors when atom-languageclient is used with TypeScript

aminya:
This fixes the type errors when atom-languageclient is used with TypeScript.

### Verification

With this PR, the following client does not error anymore: 
https://github.com/Pure-D/atomize-d/pull/34


UziTech (approved):
LGTM

PR #124, opened 2021-02-21
feat: add functions for spawning a general executable + allow custom handlers for lsProcess + fix LanguageServerProcess type

aminya:
This fills the gap for native language servers such as for D, Java, etc, or the executables that are not a JavaScript file such as for Python.

- feat: add functions for spawning a general executable: `AutolanguageClient.spawn`
- allow custom handlers for lsProcess `onSpawnClose`, `onSpawnExit`, `onSpawnError`, `onSpawnDisconnect`
- fix LanguageServerProcess type

### DIFF
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -383,6 +383,7 @@ export default class AutoLanguageClient {
   private captureServerErrors(lsProcess: LanguageServerProcess, projectPath: string): void {
     lsProcess.on('error', (err) => this.onSpawnError(err));
     lsProcess.on("close", (code, signal) => this.onSpawnClose(code, signal));
+    lsProcess.on("disconnect", () => this.onSpawnDisconnect());
     lsProcess.on('exit', (code, signal) => this.onSpawnExit(code, signal));
     lsProcess.stderr?.setEncoding('utf8');
     lsProcess.stderr?.on('data', (chunk: Buffer) => this.onSpawnStdErrData(chunk, projectPath));
@@ -412,6 +413,12 @@ export default class AutoLanguageClient {
     }
   }
 
+  /** The function called whenever the spawned server `disconnect`s.
+   *  Extend (call super.onSpawnDisconnect) or override this if you need custom disconnect handling
+   */
+  protected onSpawnDisconnect(): void {
+    this.logger.debug(`${this.getServerName()} language server for ${this.getLanguageName()} got disconnected.`);
+  }
 
   /** The function called whenever the spawned server `exit`s.
    *  Extend (call super.onSpawnExit) or override this if you need custom exit handling
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -279,6 +279,19 @@ export default class AutoLanguageClient {
     await this._serverManager.stopAllServers();
   }
 
+  /** Spawn a general language server
+   *  Use this inside the `startServerProcess` override if the language server is a general executable
+   *  Also see the `spawnChildNode` method
+   */
+  protected spawn(lspPath: string, args: string[], options: cp.SpawnOptions = {}): cp.ChildProcess {
+    this.logger.debug(`starting "${lspPath} ${args.join(' ')}"`);
+    return cp.spawn(lspPath, args, options);
+  }
+
+  /** Spawn a language server using Atom's Nodejs process
+   *  Use this inside the `startServerProcess` override if the language server is a JavaScript file.
+   *  Also see the `spawn` method
+   */
   protected spawnChildNode(args: string[], options: cp.SpawnOptions = {}): cp.ChildProcess {
     this.logger.debug(`starting child Node "${args.join(' ')}"`);
     options.env = options.env || Object.create(process.env);
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -283,9 +283,9 @@ export default class AutoLanguageClient {
    *  Use this inside the `startServerProcess` override if the language server is a general executable
    *  Also see the `spawnChildNode` method
    */
-  protected spawn(lspPath: string, args: string[], options: cp.SpawnOptions = {}): cp.ChildProcess {
-    this.logger.debug(`starting "${lspPath} ${args.join(' ')}"`);
-    return cp.spawn(lspPath, args, options);
+  protected spawn(exe: string, args: string[], options: cp.SpawnOptions = {}): cp.ChildProcess {
+    this.logger.debug(`starting "${exe} ${args.join(' ')}"`);
+    return cp.spawn(exe, args, options);
   }
 
   /** Spawn a language server using Atom's Nodejs process
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -117,9 +117,9 @@ export default class AutoLanguageClient {
   }
 
   /** (Optional) Return the parameters used to initialize a client - you may want to extend capabilities */
-  protected getInitializeParams(projectPath: string, process: LanguageServerProcess): ls.InitializeParams {
+  protected getInitializeParams(projectPath: string, lsProcess: LanguageServerProcess): ls.InitializeParams {
     return {
-      processId: process.pid,
+      processId: lsProcess.pid,
       rootPath: projectPath,
       rootUri: Convert.pathToUri(projectPath),
       workspaceFolders: null,
@@ -312,14 +312,14 @@ export default class AutoLanguageClient {
 
   /** Starts the server by starting the process, then initializing the language server and starting adapters */
   private async startServer(projectPath: string): Promise<ActiveServer> {
-    const process = await this.reportBusyWhile(
+    const lsProcess = await this.reportBusyWhile(
       `Starting ${this.getServerName()} for ${path.basename(projectPath)}`,
       async () => this.startServerProcess(projectPath),
     );
-    this.captureServerErrors(process, projectPath);
-    const connection = new LanguageClientConnection(this.createRpcConnection(process), this.logger);
+    this.captureServerErrors(lsProcess, projectPath);
+    const connection = new LanguageClientConnection(this.createRpcConnection(lsProcess), this.logger);
     this.preInitialization(connection);
-    const initializeParams = this.getInitializeParams(projectPath, process);
+    const initializeParams = this.getInitializeParams(projectPath, lsProcess);
     const initialization = connection.initialize(initializeParams);
     this.reportBusyWhile(
       `${this.getServerName()} initializing for ${path.basename(projectPath)}`,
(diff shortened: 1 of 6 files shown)


lib/auto-languageclient.ts
```diff
@@ -378,23 +392,31 @@ export default class AutoLanguageClient {
     );
   }
 
+  private handleCloseFailure(code: number | null, signal: NodeJS.Signals | null): void {
+    if (code !== 0 && signal === null) {
+      atom.notifications.addError(
+        `${this.getServerName()} language server for ${this.getLanguageName()} was closed with code: ${code}.`
```
aminya, line 398: In my opinion, this is an error, but in the above, the `exit` is considered a logger.debug

aminya: We can also add a `disconnect` listener.

lib/auto-languageclient.ts
```diff
@@ -353,11 +366,12 @@ export default class AutoLanguageClient {
...
   }
 
-  private captureServerErrors(childProcess: LanguageServerProcess, projectPath: string): void {
-    childProcess.on('error', (err) => this.handleSpawnFailure(err));
-    childProcess.on('exit', (code, signal) => this.logger.debug(`exit: code ${code} signal ${signal}`));
-    childProcess.stderr.setEncoding('utf8');
-    childProcess.stderr.on('data', (chunk: Buffer) => {
+  private captureServerErrors(lsProcess: LanguageServerProcess, projectPath: string): void {
+    lsProcess.on('error', (err) => this.handleSpawnFailure(err));
+    lsProcess.on("close", (...args) => this.handleCloseFailure(...args));
```
aminya, line 371: We can also add a `disconnect` listener.

lib/utils.ts
```diff
@@ -111,3 +112,18 @@ export function promiseWithTimeout<T>(ms: number, promise: Promise<T>): Promise<
...
+ * @param exeName name of the exe file
+ * @param rootPath the path of the folder of the exe file. Defaults to 'bin'
+ * @param exeExtention the extention of the exe file. Defaults to `process.platform === "win32" ? ".exe" : ""`
+ */
+export function getExePath(
+ exeName: string,
+ rootPath = "bin",
+ exeExtention = process.platform === "win32" ? ".exe" : ""
+): string {
+ return resolve(join(rootPath, `${process.platform}-${process.arch}`, `${exeName}${exeExtention}`));
```
UziTech, line 128: should `` `${process.platform}-${process.arch}` `` be configurable? maybe as part of `rootPath` like

```diff
- rootPath = "bin",
+ rootPath = join("bin", `${process.platform}-${process.arch}`),
```

UziTech: I almost feel like this doesn't need to be a utility function since it isn't much easier than writing 

```js
resolve(join(rootPath, `${process.platform}-${process.arch}`, `serve-d${process.platform === "win32" ? ".exe" : ""}`))
```

And that is infinitely more configurable than this utility function.

aminya: Yes. Fixed in https://github.com/atom-community/atom-languageclient/pull/124/commits/9746992e9a358815e861cbc48753e52d74fe45c1

lib/utils.ts
```diff
@@ -111,3 +112,18 @@ export function promiseWithTimeout<T>(ms: number, promise: Promise<T>): Promise<
...
+ * @param rootPath the path of the folder of the exe file. Defaults to 'join("bin", `${process.platform}-${process.arch}`)'
+ * @param exeExtention the extention of the exe file. Defaults to `process.platform === "win32" ? ".exe" : ""`
+ */
+export function getExePath(
+ exeName: string,
+ rootPath = join("bin", `${process.platform}-${process.arch}`),
+ exeExtention = process.platform === "win32" ? ".exe" : ""
+): string {
+ return resolve(join(rootPath, `${exeName}${exeExtention}`));
+}
```
UziTech, line 129: What if the file has the same executable for multiple platforms? would they have to copy the executable into multiple folders?

This utility function doesn't seem very useful since it is abstracting a one liner that is much more configurable.

aminya: If the exe is the same, they can just pass it directly to spawn.

UziTech: > This is useful for native exe files, in which the exe is different for different for each platform and arch.

so only useful for very specific situation. Doesn't seem like a good candidate for a utility function.

aminya: This is not "over specific". Rust, D, Go, C++, C, etc all have a similar situation.

UziTech: > In contrast, assuming that everything is a Node LSP is "over specific".

Nobody assumes that. It is just that writing `cp.spawn(args)` is just as easy as `super.spawn(args)` so it doesn't seem to make sense to abstract it.

aminya: Well, I was stuck on this for two weeks. I thought that `spawnChildNode` is the way to use it for all the LSP stuff. Until I realized that is far from the truth.

Having some structure makes it easier to scale this and get reproducible results at scale. Not having anything considered gives the wrong signal.


For the same reason, that's why everyone uses `prebuildify`. This unified structure has made it easier to provide native addons at scale.  If you want to look at the issue from your point of view, `prebuildify` simply calls `prebuild` and copies the binaries to a predefined folder.

UziTech: Having the documentation on how to use an executable file would be good but if we want to abstract it for the `prebuildify` workflow that would be fine.

something like:

```js
super.spawnPrebuildify(
      'serve-d',
      args,
      opts
    )
```

anyone who wants more customizability can just us `cp.spawn`

aminya: The current implementation I have chosen covers all the cases:
- native exe by using `getExePath` (useful for native languages such as D, C++)
- manual management
We already had Nodejs spawning which in combination with the above covers all the cases.

I don't want to become religious, but I think this provides a clean way to simplify the ide packages we already have. Many of the ide packages (e.g. ide-python) do manual error management for the spawned process because of the wrong signal this missing method had given, even though we already manage it internally!

UziTech: I'm not totally against abstracting `spawn` to make the "usual" process easier. It is mostly `getExePath` that seems like a waist of an abstraction.

aminya: Here, I am in favor of providing small functions to allow the developers to compose them as they wish. This is cleaner to me in this specific case. 

But if you really want, I can merge `getExePath` into the `spawn` method. In my opinion, the interface is already clean, and merging `getExePath` can make it hard to use.

aminya: I found a way to extend the functionality of `getExePath` to the cases in which the name is on the `PATH`

UziTech: > merging  `getExePath` can make it hard to use.

How can it make it hard to use? if they use prebuildify (or the same directory structure) they should be able to use the default configuration. (Which I'm assuming is most language servers.)

And if they do not use the same directory structure for the binaries it is just as simple to use `cp.spawn(serverExe, ...)` as it is to use `super.spawn(...)`

Esentially if the package uses the "usual" structure we make it easier. If they don't then it is the same as before.

UziTech: > Here, I am in favor of providing small functions to allow the developers to compose them as they wish.

For a general library like underscore, I agree, but this library is for very specific lsp atom packages.

UziTech: > I found a way to extend the functionality of `getExePath` to the cases in which the name is on the `PATH`

Shouldn't the package that is using `getExePath` know whether the exe is provided or it should return the `PATH`?

aminya: > Shouldn't the package that is using getExePath know whether the exe is provided or it should be in the PATH?

Certainly if they bundle it with the package, but in another case, the file might not exist there. For example, you can say generally Python is installed in `~/usr/bin`, but it might not be there, and so you can use the exe on the PATH.

UziTech: Ya so if they know python is not bundled why would they use `exePath = getExePath("python")` instead of `exePath = "python"`?

UziTech: `getExePath` looks like a solution without a problem.

UziTech: Is `getExePath`  useful to lsp packages somewhere other than starting the server? If not, the code should be merged into the `spawn` function.

aminya: Merged them in https://github.com/atom-community/atom-languageclient/pull/124/commits/a5c03a572bacd1a12eca83fbced80be942cc5ed6

lib/auto-languageclient.ts
```diff
@@ -353,22 +380,19 @@ export default class AutoLanguageClient {
...
-      this.processStdErr = (this.processStdErr + errorString)
-        .split('\n')
-        .slice(-5)
-        .join('\n');
-    });
+  private captureServerErrors(lsProcess: LanguageServerProcess, projectPath: string): void {
+    lsProcess.on('error', (err) => this.onSpawnError(err));
+    lsProcess.on("close", (code, signal) => this.onSpawnClose(code, signal));
+    lsProcess.on("disconnect", () => this.onSpawnDisconnect());
+    lsProcess.on('exit', (code, signal) => this.onSpawnExit(code, signal));
```
UziTech, line 387: A little nit picky but mixing double and single quotes. Maybe after this is merged we should add prettier or a linting rule.

lib/auto-languageclient.ts
```diff
@@ -279,7 +279,34 @@ export default class AutoLanguageClient {
...
+   * Also see the `spawnChildNode` method.
+   * If the name is provided as the first argument, it checks `bin/platform-arch/exeName` by default, and if doesn't exists uses the exe on PATH.
+   * For example on Windows x64, by passing `serve-d`, `bin/win32-x64/exeName.exe` is spawned by default.
+   * @param exe the `name` or `path` of the executable
+   * @param args args passed to spawn the exe. Defaults to `[]`.
+   * @param options: child process spawn options. Defaults to `{}`.
+   * @param rootPath the path of the folder of the exe file. Defaults to `join("bin", `${process.platform}-${process.arch}`)`.
+   * @param exeExtention the extention of the exe file. Defaults to `process.platform === "win32" ? ".exe" : ""`
+   */
+  protected spawn(
```
UziTech, line 294: Should this be called `spawnChildExe` to compliment `spawnChildNode`?

lib/auto-languageclient.ts
UziTech: Ya I would think this should be an error by default as long as there is a way for packages to override it if it shouldn't be an error for them.

UziTech (commented):
A couple nits otherwise LGTM

lib/auto-languageclient.ts
aminya: Ah right. It is hard to maintain the style without Prettier. I will make a new PR after this fixing the format.

lib/auto-languageclient.ts
aminya: I think `spawnChildNode` comes from the fact that Atom's Node process (a real child) is used for spawning. But `spawn` is general (the spawned exe is independent), and using `Child` can be misleading.

lib/auto-languageclient.ts
aminya: They can do this by overriding the method.

aminya:
I will create a prettier PR after this. Also, as I explained `spawn` is a more suitable name because the exe is independent of Node.

PR #125, opened 2021-02-23
chore: add prettier and use eslint-config-atomic

### LINKED ISSUE #115: Set up Prettier

aminya:
Fixes #115 

The open PRs need to be updated by running `pnpm format`

### DIFF
--- a/lib/convert.ts
+++ b/lib/convert.ts
@@ -1,12 +1,7 @@
 import type * as atomIde from 'atom-ide-base';
 import * as ls from './languageclient';
 import * as URL from 'url';
-import {
-  Point,
-  FilesystemChange,
-  Range,
-  TextEditor,
-} from 'atom';
+import { Point, FilesystemChange, Range, TextEditor } from 'atom';
 
 /**
  * Public: Class that contains a number of helper methods for general conversions
@@ -115,10 +110,7 @@ export default class Convert {
    * @returns A {TextDocumentPositionParams} that has textDocument property with the editors {TextDocumentIdentifier}
    *   and a position property with the supplied point (or current cursor position when not specified).
    */
-  public static editorToTextDocumentPositionParams(
-    editor: TextEditor,
-    point?: Point,
-  ): ls.TextDocumentPositionParams {
+  public static editorToTextDocumentPositionParams(editor: TextEditor, point?: Point): ls.TextDocumentPositionParams {
     return {
       textDocument: Convert.editorToTextDocumentIdentifier(editor),
       position: Convert.pointToPosition(point != null ? point : editor.getCursorBufferPosition()),
@@ -177,7 +169,7 @@ export default class Convert {
       case 'deleted':
         return [{ uri: Convert.pathToUri(fileEvent.path), type: ls.FileChangeType.Deleted }];
       case 'renamed': {
-        const results: Array<{ uri: string, type: ls.FileChangeType }> = [];
+        const results: Array<{ uri: string; type: ls.FileChangeType }> = [];
         if (fileEvent.oldPath) {
           results.push({ uri: Convert.pathToUri(fileEvent.oldPath), type: ls.FileChangeType.Deleted });
         }
--- a/lib/convert.ts
+++ b/lib/convert.ts
@@ -1,6 +1,6 @@
 import type * as atomIde from "atom-ide-base"
 import * as ls from "./languageclient"
-import * as URL from "url"
+import { URL } from "url"
 import { Point, FilesystemChange, Range, TextEditor } from "atom"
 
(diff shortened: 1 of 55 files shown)


prettier.config.js
```diff
@@ -0,0 +1,25 @@
+// Add to .prettierignore to ignore files and folders
+
+// This configuration all the formats including typescript, javascript, json, yaml, markdown
+module.exports = {
+  tabWidth: 2,
+  printWidth: 120,
+  semi: false,
+  singleQuote: false,
```
aminya, line 8: To reduce the code diff we can modify these, but that means changing the preset that the organization uses.

lib/utils.ts
```diff
@@ -24,7 +24,7 @@ export function getWordAtPosition(editor: TextEditor, position: Point): Range {
 
 export function escapeRegExp(string: string): string {
   // From atom/underscore-plus.
-  return string.replace(/[-/\\^$*+?.()|[\]{}]/g, "\\$&")
+  return string.replace(/[$()*+./?[\\\]^{|}-]/g, "\\$&")
```
aminya, line 27: This change seems correct, but we can discuss it more.

prettier.config.js
UziTech: I would say use organization preset. I would rather have consistency than smaller diff.

lib/utils.ts
UziTech: It is the same but odd that prettier changes it.

prettier.config.js
aminya: OK. I will revert this then.

lib/utils.ts
aminya: It is not prettier, it is the optimize-regex plugin for Eslint. I recently added it to the Eslint config

lib/convert.ts
```diff
@@ -36,20 +31,20 @@ export default class Convert {
    *   to deal with http/https sources in the future.
    */
   public static uriToPath(uri: string): string {
-    const url = URL.parse(uri);
```
aminya, line 39: URL.parse is deprecated

UziTech:
Can you add prettier to the lint check in github actions. And the lint warnings don't seem to be working. I don't see any messages about where linting is failing but there are a lot of warnings in the lint check.

lib/convert.ts
aminya: @UziTech If we fix this, we can enable the strict variant of the config

aminya:
> I still think they should be errors so linting will fail if someone submits a PR that doesn't pass linting.

Check this: https://github.com/atom-community/atom-languageclient/pull/125#discussion_r580733035

lib/convert.ts
UziTech: Can't we just change it to `new URL(uri)`?

aminya: `url.path` does not exist in the new API. There is `pathname`, but I am not sure if it is the same.

UziTech: according to the [docs](https://nodejs.org/api/url.html#url_urlobject_path): `The path property is a concatenation of the pathname and search components.`

aminya: So, I think it is fixed in 
https://github.com/atom-community/atom-languageclient/pull/125/commits/358d254391847077309e9e1170156a94e8d4f27a

aminya: Strangely, the tests only pass on Windows, and they fail on the Unix systems

UziTech: I say we just add `/* eslint-disable node/no-deprecated-api */` at the top of the file for now and fix it in another PR.

UziTech: It seems like it is failing because of [this](https://stackoverflow.com/questions/59375013/node-legacy-url-parse-deprecated-what-to-use-instead) so we will have to figure out some other fix than just using the new URL api.

aminya: I made the error a warning for now, so we can merge this PR.

UziTech (approved):

PR #126, opened 2021-02-23
fix: remove deprecated url.parse

UziTech:


test/convert.test.ts
```diff
@@ -81,7 +81,7 @@ describe("Convert", () => {
 
     it("parses URI without double slash in the beginning on Windows", () => {
       setProcessPlatform("win32")
-      expect(Convert.uriToPath("file:/x:/a/b/c/d.txt")).equals("x:\\a\\b\\c\\d.txt")
+      expect(Convert.uriToPath("file:/x:/a/b/c/d.txt").toLowerCase()).equals("x:\\a\\b\\c\\d.txt")
```
UziTech, line 84: Apparently the `URL` api uppercases drive letters on windows but I don't think that would actually be an issue since windows file paths are case insensitive.

lib/convert.ts
```diff
@@ -31,12 +30,12 @@ export default class Convert {
    *   to deal with http/https sources in the future.
    */
   public static uriToPath(uri: string): string {
-    const url = URL.parse(uri) /*eslint node/no-deprecated-api: "warn"*/ // TODO
-    if (url.protocol !== "file:" || url.path === undefined || url.path === null) {
+    const url = new URL(uri, "file://")
```
UziTech, line 33: The trick was adding `file://` as the base url.

lib/convert.ts
```diff
@@ -31,12 +30,12 @@ export default class Convert {
    *   to deal with http/https sources in the future.
    */
   public static uriToPath(uri: string): string {
-    const url = URL.parse(uri) /*eslint node/no-deprecated-api: "warn"*/ // TODO
-    if (url.protocol !== "file:" || url.path === undefined || url.path === null) {
```
aminya, line 35: Does it become null ever?

test/convert.test.ts
aminya: Does it work on other operating systems as expected?

UziTech: the issue is with the test. Other OSes keep the given case but windows uppercases the drive letters.

Really this test should only run on windows instead of changing the process to look like it is on windows.

lib/convert.ts
UziTech: I don't think so but it doesn't hurt to have the check.

aminya (approved):

PR #129, opened 2021-03-14
Feature: handling custom requests

laughedelic:
Ported https://github.com/atom/atom-languageclient/pull/249 to the new repo.

- added missing `onCustomRequest` method
- renamed `onCustom` to `onCustomNotification`

### DIFF
--- a/lib/languageclient.ts
+++ b/lib/languageclient.ts
@@ -102,7 +102,9 @@ export class LanguageClientConnection extends EventEmitter {
     this._onNotification({ method }, callback)
   }
 
-  // @deprecated Use onCustomNotification method instead
+  /**
+   * @deprecated Use `onCustomNotification` method instead
+   */
   public onCustom(method: string, callback: (obj: object) => void): void {
     this.onCustomNotification(method, callback)
   }
--- a/lib/languageclient.ts
+++ b/lib/languageclient.ts
@@ -17,6 +17,7 @@ export interface KnownNotifications {
 export interface KnownRequests {
   "window/showMessageRequest": [lsp.ShowMessageRequestParams, lsp.MessageActionItem | null]
   "workspace/applyEdit": [lsp.ApplyWorkspaceEditParams, lsp.ApplyWorkspaceEditResponse]
+  [custom: string]: [object, object | null]
 }
 
 export type RequestCallback<T extends keyof KnownRequests> = KnownRequests[T] extends [infer U, infer V]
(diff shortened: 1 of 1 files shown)


lib/languageclient.ts
```diff
@@ -91,16 +92,32 @@ export class LanguageClientConnection extends EventEmitter {
...
    */
-  public onCustom(method: string, callback: (obj: object) => void): void {
+  public onCustomNotification(method: string, callback: (obj: object) => void): void {
     this._onNotification({ method }, callback)
   }
 
+  // @deprecated Use onCustomNotification method instead
+  public onCustom(method: string, callback: (obj: object) => void): void {
+    this.onCustomNotification(method, callback)
+  }
```
laughedelic, line 108: I'm not sure this is the right way to deprecate in TypeScript and probably it's too harsh to deprecate it at all. I just think `onCustom` name is confusing, because it's only about notifications.

lib/languageclient.ts
```diff
@@ -91,16 +92,32 @@ export class LanguageClientConnection extends EventEmitter {
...
+
+  /**
+   * Public: Register a callback for a custom request
+   *
+   * @param method A string containing the name of the message to listen for.
+   * @param callback The function to be called when the message is received.
+   *   The payload from the message is passed to the function.
+   */
+  public onCustomRequest(method: string, callback: (obj: object) => Promise<object>): void {
```
laughedelic, line 117: Is it ok or the type should be `callback: (obj: object) => Promise<object | void>`?

lib/languageclient.ts
UziTech: I think it requires a doc comment.

```suggestion
  /**
   * @deprecated Use `onCustomNotification` method instead
   */
  public onCustom(method: string, callback: (obj: object) => void): void {
    this.onCustomNotification(method, callback)
  }
```

aminya:
The CI failure is fixed in #131

lib/languageclient.ts
aminya: If the return type is not needed all the time, it is OK to change it to `object | void`.

lib/languageclient.ts
```diff
@@ -17,6 +17,7 @@ export interface KnownNotifications {
 export interface KnownRequests {
   "window/showMessageRequest": [lsp.ShowMessageRequestParams, lsp.MessageActionItem | null]
   "workspace/applyEdit": [lsp.ApplyWorkspaceEditParams, lsp.ApplyWorkspaceEditResponse]
+  [custom: string]: [object, object | null]
```
aminya, line 20:
```suggestion
  Record<string, [object, object | null]>
```

aminya:
@laughedelic Could you take a look at the suggestions?

laughedelic:
@aminya yes, thanks for the review! I'll try to address it tonight.

aminya:
@laughedelic are you able to reword the commits to start with `fix: a lowercase word`? If you are away from the keyboard, I can do that.

laughedelic:
@aminya is it ok if I just squash them into one with the right format?

aminya:
I'd prefer separate commits because it gives us more flexibility later for a potential reverting.

You can use the following to reword all the commits one by one:

```
git rebase -i 488f0033e8e8698a68becce2d78f35745318545e
```

Then find and replace `pick` with `r` in the opened document, and reword, save the document, close one by one.

aminya (approved):
Thank you!

PR #130, opened 2021-03-15
test: simplify autocomplete-paths tests by removing createCompletionItem

aminya:
### Description

This removes the `createCompletionItem` function from the tests and just uses `CompletionItem` directly. This simplifies the tests, so we no longer need to update this function in two places as LSP changes.

`CompletionItem` requires label property which is added explicitly in the tests.


aminya:
The CI failure is fixed in #131

UziTech (approved):

PR #133, opened 2021-03-16
refactor: clean-up module augments

lierdakil:
This is still not pretty, but arguably less painful than those `*Ext` declarations.

I can't spend much time now on searching through the TypeScript issue tracker and/or constructing a minimal testcase for the bugreport, but we should at least check if it's been already reported at some point.

### DIFF
--- a/typings/atom/index.d.ts
+++ b/typings/atom/index.d.ts
@@ -1,12 +1,19 @@
 import { Point, Notification, NotificationOptions, TextEditor } from "atom"
 
-declare module "atom" {
-  interface TextEditorExt extends TextEditor {
+// NOTE: due to a bug in how TypeScript resolves ambient augments,
+// need to be more specific here for TextEditor to keep its "class"
+// status and not be demoted to "interface". Should revert this
+// once the issue is fixed.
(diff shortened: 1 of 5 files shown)


aminya (changes requested):
Could you update @types/atom as well?

typings/atom/index.d.ts
```diff
@@ -1,12 +1,19 @@
 import { Point, Notification, NotificationOptions, TextEditor } from "atom"
```
aminya, line 2: The lint error might go away by using `export {}` or exporting the interfaces

lierdakil: But... the lint is plain wrong, isn't it?

lierdakil: After exporting, eslint complains about name shadowing, which is just stupid. It doesn't seem to understand what module augmentations are.

lierdakil: There are many, many other eslint failures here. Fixing those is out of scope.

aminya (approved):
Thanks

PR #135, opened 2021-03-16
feat: implement window/showDocument

aminya:
Implements `ShowDocumentAdapter`, which adapts the window/showDocument command to Atom's text editors or external programs.

Extracted from #112


lib/adapters/show-document-adapter.ts
```diff
@@ -0,0 +1,41 @@
...
+        initialColumn: params.selection?.start.character ?? 0,
+      })
+      if (view instanceof TextEditor && params.selection != null) {
+        view.selectToBufferPosition(Convert.positionToPoint(params.selection.end))
+      }
+      return { success: true }
+    } else {
+      // open using Electron
+      shell.openExternal(params.uri, { activate: params.takeFocus })
+      return { success: false }
```
aminya, line 38:
```suggestion
      return { success: true }
```

test/adapters/show-document-adapter.test.ts
```diff
@@ -0,0 +1,22 @@
...
+describe("ShowDocumentAdapter", () => {
+  describe("can attach to a server", () => {
+    const lcc = new LanguageClientConnection(createSpyConnection())
+
+    it("subscribes to onShowDocument", async () => {
+      const spy = sinon.spy()
+      lcc["_onRequest"] = spy
+
+      ShowDocumentAdapter.attach(lcc)
+      expect((lcc["_onRequest"] as sinon.SinonSpy).calledOnce).to.be.true
```
UziTech, line 16: I feel like this is a good example of testing the implementation (which is a bad practice). We should be testing that the correct arguments are sent to the rpc connection.

UziTech (changes requested):

test/adapters/show-document-adapter.test.ts
aminya: `attach` function is very simple. This is all we want. I will add tests for the actual `showDocument`

UziTech: But this still isn't testing the correct thing. `_onRequest` is a private method so if we ever decide to remove it and inline it's function wherever it is used we shouldn't have to change any tests since it is an implementation detail.

The correct test should be something like:

```
  describe("can attach to a server", () => {
    const connection = createSpyConnection()
    const lcc = new LanguageClientConnection(connection)

    it("subscribes to onShowDocument", async () => {
      const spy = sinon.spy()
			ShowDocumentAdapter.showDocument = spy
      ShowDocumentAdapter.attach(lcc)
      expect((connection.onRequest as sinon.SinonSpy).calledOnce).to.be.true
      const spyArgs = connection.onRequest.firstCall.args
      expect(spyArgs[0]).to.deep.equal({ method: "window/showDocument" })
			spyArgs[1]("test")
      expect((ShowDocumentAdapter.showDocument as sinon.SinonSpy).calledOnce).to.be.true
      expect(ShowDocumentAdapter.showDocument.firstCall.args).to.equal("test")
    })
  })
})
```

UziTech: Or we would at least have to test that `_onRequest` actually does what it is supposed to as well. Right now we are only testing the attach function half way.

aminya: Fixed in https://github.com/atom-community/atom-languageclient/pull/135/commits/43e195dfd551278a2970c19bf6266a9b4cf74fb5

Yes, we can add tests for `_onRequest`. Because this PR doesn't edit that, the test for it is out of the scope of this PR.

lib/adapters/show-document-adapter.ts
```diff
@@ -0,0 +1,52 @@
+import { shell } from "electron"
+import { LanguageClientConnection, ShowDocumentParams, ShowDocumentResult } from "../languageclient"
+import { TextEditor } from "atom"
+import Convert from "../convert"
+
+/** Public: Adapts the window/showDocument command to Atom's text editors or external programs. */
+export class ShowDocumentAdapter {
```
UziTech, line 7: What is the reason we export a class and the functions separately?

UziTech: Can we just make the functions static like other adapters.

aminya: I added the class just in case someone wants to overload them. Otherwise, I would remove it.

aminya: I think I can remove the class. Since people can call the functions directly. There is no point in adding class bloat here.

aminya:
The tests fail for different reasons on Windows vs Unix. Not sure if my uri for opening the fixture is correct

UziTech:
apparently `atom.workspace.open` does different things when given an invalid path on windows and linux.

aminya:
> apparently `atom.workspace.open` does different things when given an invalid path on windows and linux.

That's a bug that should be taken care of inside Atom itself.

aminya (commented):
This is ready to go

test/adapters/show-document-adapter.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      })
+
+      it("selects the given selection range", async () => {
+        const selectionLSRange = { start: { line: 0, character: 31 }, end: { line: 0, character: 44 } }
+        const params: ShowDocumentParams = {
+          uri: helloPath,
+          selection: selectionLSRange,
+        }
+        const editor = await canShowDocumentInAtom(params)
+        expect(editor.getSelectedBufferRange()).deep.equal(Convert.lsRangeToAtomRange(selectionLSRange))
```
aminya, line 89: Not sure why it fails

aminya: Never mind. The JavaScript file generated by TypeScript adds `"use strict"` on top.

Fixed

lib/adapters/show-document-adapter.ts
```diff
@@ -0,0 +1,59 @@
...
+ * {@inheritDoc ShowDocumentParams}
+ */
+export async function showDocument(params: ShowDocumentParams): Promise<ShowDocumentResult> {
+  try {
+    if (!params.external) {
+      // open using atom.workspace
+      const view = await atom.workspace.open(Convert.uriToPath(params.uri), {
+        activateItem: params.takeFocus,
+        activatePane: params.takeFocus,
+        pending: true,
```
UziTech, line 40: Is there a reason this is opened as a pending item? (Not saying it shouldn't, just wondering if that is part of the spec?)

lib/adapters/show-document-adapter.ts
```diff
@@ -0,0 +1,59 @@
...
+        initialLine: params.selection?.start.line ?? 0,
+        initialColumn: params.selection?.start.character ?? 0,
+      })
+      if (view === undefined) {
```
UziTech, line 44: I'm not convinced that this won't return null in the future. It should always be truthy if it exists.

```suggestion
      if (!view) {
```

lib/adapters/show-document-adapter.ts
aminya: This is what it means. Opening in a pending is an Atom thing and not part of the spec.
```ts
    /**
     *  A Boolean indicating whether or not the item should be opened in a pending state.
     *  Existing pending items in a pane are replaced with new pending items when they
     *  are opened.
     */
```

lib/adapters/show-document-adapter.ts
UziTech: We should probably keep it consistent with the other adapters. I'm ok with removing the class if that would work for the others. (Probably in another PR)

lib/adapters/show-document-adapter.ts
UziTech: I know what it is. I'm not sure if that is actually what the user would want. Or if there is some way we could make that configurable.

UziTech: Maybe we should set it as the value of `atom.config.get('core.allowPendingPaneItems')`

UziTech: Actually it looks like Atom ignores it if that setting is set anyway.

https://github.com/atom/atom/blob/757478442696b302bb8f9ad16e93b30f9350d949/src/workspace.js#L1102

aminya: In practice, I don't see any difference. I am not sure what it exactly does. Maybe better to improve the documentation for the `pending` option before changing the code here.

lib/adapters/show-document-adapter.ts
```diff
@@ -0,0 +1,59 @@
+import { shell } from "electron"
+import { LanguageClientConnection, ShowDocumentParams, ShowDocumentResult } from "../languageclient"
+import { TextEditor } from "atom"
+import Convert from "../convert"
+
+/** Public: Adapts the window/showDocument command to Atom's text editors or external programs. */
+export default class ShowDocumentAdapter {
+  /** {@inheritDoc attach} */
+  public attach(arg: Parameters<typeof attach>[0]): ReturnType<typeof attach> {
```
UziTech, line 9: Should these functions be static?

lib/adapters/show-document-adapter.ts
UziTech: After looking at the other adapters it seems like exporting the class with static functions would be better to keep importing adapters consistent. Otherwise it will be a pain figuring out which adapters need `import * as adapter from 'adapter'` and which ones need `import adapter from 'adapter'`. It would be better to only import on way.

lib/adapters/show-document-adapter.ts
aminya: They can, but in my opinion, static methods are fundamentally wrong, because they do not use anything from the class, and the class is just being misused as a namespace!

To be honest, I have just kept this class to be consistent with other adapters, otherwise, I would have removed it.

UziTech: > because they do not use anything from the class, and the class is just being misused as a namespace!

I think that is kind of the point. These function don't use anything from the class. The class is just a namespace. I don't see what the problem is with that.

I feel like something like:
```js
import ShowDocumentAdapter from '...'
...
ShowDocumentAdapter.attach(...)
```

is better than:

```js
import {attach} from '...'
...
attach(...) // where does this non-namespaced attach come from?
```

Namespacing is very useful.

aminya: I don't say that namespacing is not useful, but JavaScript and TypeScript both support namespaces, and there is no need to use a class as a namespace.
```ts
import * as ns from "...."
ns.attach()
```
or
```js
const ns = require("...")
ns.attach()
```

aminya: I found an easier way to maintain consistency.
https://github.com/atom-community/atom-languageclient/pull/135/commits/96926433a19296e94e7df49cddc72119edced873

UziTech: I still think these should be consistent with other adapters.

The biggest reason is so we don't have to figure out if it is supposed to be `import ... from ...` or `import * as ... from ...` Which one works for each adapter? It would be better if they were all the same.

aminya: The way I have `export default` here, make `import ... from ...` working like other adapters.

UziTech: > there is no need to use a class as a namespace.

I could argue that there is no need to *not* use a class as a namespace. Is there a good reason to maintain both ways? I feel like it is just extra maintenance without any benefit.

aminya: This is just a simple `export default { export1, export2 } `. There is no maintenance cost involved with re-exporting.

UziTech: What you just said is extra maintenance by definition. Not a lot of extra maintenance, granted, but still extra maintenance for no benefit. 

I guess we can merge it this way and maybe make it consistent in another PR down the road.

UziTech (approved):

PR #136, opened 2021-03-19
Autocomplete improvements

aminya:
Extracted from #112

### DIFF
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -551,6 +551,7 @@ export default class AutoLanguageClient {
       inclusionPriority: 1,
       suggestionPriority: 2,
       excludeLowerPriority: false,
+      filterSuggestions: true,
       getSuggestions: this.getSuggestions.bind(this),
       onDidInsertSuggestion: this.onDidInsertSuggestion.bind(this),
       getSuggestionDetailsOnSelect: this.getSuggestionDetailsOnSelect.bind(this),
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -517,11 +517,23 @@ export default class AutoLanguageClient {
   }
 
   // Autocomplete+ via LS completion---------------------------------------
+
+  /**
+   * A method to override to return an array of grammar scopes that should not be used for autocompletion.
+   *
+   * Usually that's used for disabling autocomplete inside comments,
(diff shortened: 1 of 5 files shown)


lib/auto-languageclient.ts
```diff
@@ -517,11 +533,21 @@ export default class AutoLanguageClient {
...
+  protected getAutocompleteDisabledScopes(): Array<string> {
+    return []
+  }
+
   public provideAutocomplete(): ac.AutocompleteProvider {
     return {
-      selector: this.getGrammarScopes()
-        .map((g) => (g.includes(".") ? "." + g : g))
-        .join(", "),
+      selector: this.getGrammarScopes().join(", "),
```
aminya, line 549: Not so sure about this change. 

cc: @illright

illright: Well, scopes are defined as class strings, this behaviour was sort of excusing the misuse of grammar scopes. It would allow stuff like `[ 'source.python' ]`, while actual scopes should be specified as `[ '.source.python' ]`. I guess this change is kind of breaking, but then again, the breaking would be a result of incorrectly using the scopes.

aminya: Reverted in https://github.com/atom-community/atom-languageclient/pull/136/commits/cc6b7bf2912fe712115e2ffd3e70d812230cce31

lib/utils.ts
```diff
@@ -123,3 +123,12 @@ export function getExePath(exeName: string, rootPath = rootPathDefault, exeExten
...
   }
 }
+
+/**
+ * Normalizes the given grammar scope so it always starts with `.`
+ * @param grammarScope such as 'source.python' or '.source.python'
+ * @returns the normalized grammarScope such as `.source.python`
+ */
+export function normalizeGrammarScope(grammarScope: string): string {
+  return grammarScope.includes(".") && grammarScope[0] !== "." ? `.${grammarScope}` : grammarScope
```
aminya, line 133: My `normalizeGrammarScope` is a little different than what was being used here. We should determine the desired output and then normalize for that.

aminya: Not sure why autocomplete-plug needs dots in the beginning. Probably we should add tests to prevent accidental breakage in the future.

UziTech (approved):

PR #139, opened 2021-03-19
feat: initial code actions backend for Linter service 

aminya:
- Moves the old atom-ide-diagnostics code to a separate adapter
- Adds initial code actions support for the Linter package -> The packages themselves have not implemented this yet

### DIFF
--- a/lib/adapters/code-action-adapter.ts
+++ b/lib/adapters/code-action-adapter.ts
@@ -103,7 +103,7 @@ function createCodeActionParams(
     diagnostics = []
   } else {
     // TODO compile time dispatch using function names
-    diagnostics = isLinterMessage(linterMessages)
+    diagnostics = areLinterMessages(linterMessages)
       ? linterAdapter.getLSDiagnosticsForMessages(linterMessages as linter.Message[])
       : (linterAdapter as IdeDiagnosticAdapter).getLSDiagnosticsForIdeDiagnostics(
           linterMessages as atomIde.Diagnostic[],
@@ -119,8 +119,8 @@ function createCodeActionParams(
   }
 }
 
-function isLinterMessage(linterMessage: linter.Message[] | atomIde.Diagnostic[]): boolean {
-  if (linterMessage.length > 0 && "excerpt" in linterMessage[0]) {
+function areLinterMessages(linterMessages: linter.Message[] | atomIde.Diagnostic[]): boolean {
+  if ("excerpt" in linterMessages[0]) {
     return true
   }
   return false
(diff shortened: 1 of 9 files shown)


lib/adapters/code-action-adapter.ts
```diff
@@ -81,34 +82,50 @@ export default class CodeActionAdapter {
...
     }
   }
+}
 
-  private static createCodeActionParams(
+function createCodeActionParams(
     linterAdapter: LinterPushV2Adapter,
     editor: TextEditor,
     range: Range,
     diagnostics: atomIde.Diagnostic[]
```
aminya, line 97: I think we need to break the old `atomIde.Diagnostic` in favor of `Linter.Message` here.

aminya: No. We can provide service for both in AutoLanguageClient. We just need to call the correct function.

lib/adapters/linter-push-v2-adapter.ts
```diff
@@ -111,25 +144,128 @@ export default class LinterPushV2Adapter {
...
+      position: Convert.lsRangeToAtomRange(diagnostic.range),
+    },
+    reference: relatedInformationToReference(diagnostic.relatedInformation),
+    url: diagnostic.codeDescription?.href,
+    icon: iconForLSSeverity(diagnostic.severity ?? DiagnosticSeverity.Error),
+    excerpt: diagnostic.message,
+    linterName: diagnostic.source,
+    severity: lsSeverityToV2MessageSeverity(diagnostic.severity ?? DiagnosticSeverity.Error),
+    // BLOCKED: on steelbrain/linter#1722
+    solutions: undefined,
```
aminya, line 170: Not sure how to provide solutions from `Ls.Diagnostic`

bennypowers:
Can you show us a screenshot of the UI?

are there new commands like "show solutions under cursor" or "apply solutions" or something?

aminya:
> Can you show us a screenshot of the UI?
> 
> are there new commands like "show solutions under cursor" or "apply solutions" or something?

This is the backend adapter for code actions. I haven't tested its UI yet. 

We need to fix these two issues to support the UI as well
https://github.com/steelbrain/linter-ui-default/pull/611
https://github.com/steelbrain/linter/pull/1722

aminya (commented):
This can be merged as of its current state because it is backward compatible and the changes are tested. We should update it later again once Linter and Linter-ui-default actually implemented the UI for this. We can't do anything else here for now.

Because the changes edit multiple files, we should merge this first before we can fix #156

UziTech (approved):
LGTM

PR #140, opened 2021-04-10
feat: support resource operations

cdaguerre:
This is my first experience with typescript, feedback much appreciated ;)


UziTech:
Can you give some explanation of why this is needed? Examples of what this enables that doesn't currently work?

cdaguerre:
@UziTech Sure!

This enables resource operations (basically file and directory operations) as part of `workspaceEdit`s, which is currently not implemented at all.
I came across this while working on the `rename` functionality for the [ide-intelephense](https://github.com/cdaguerre/ide-intelephense) package. If this PR makes it, i'll open another one to improve the rename functionality aswell (which to my understanding doesn't work at all right now).

From https://microsoft.github.io/language-server-protocol/specification#workspaceEdit:
> Since version 3.13.0 a workspace edit can contain resource operations (create, delete or rename files and folders) as well. If resource operations are present clients need to execute the operations in the order in which they are provided. So a workspace edit for example can consist of the following two changes: (1) create file a.txt and (2) a text document edit which insert text into file a.txt.
And:
```typescript
export interface WorkspaceEdit {
	/**
	 * Holds changes to existing resources.
	 */
	changes?: { [uri: DocumentUri]: TextEdit[]; };

	/**
	 * Depending on the client capability
	 * `workspace.workspaceEdit.resourceOperations` document changes are either
	 * an array of `TextDocumentEdit`s to express changes to n different text
	 * documents where each text document edit addresses a specific version of
	 * a text document. Or it can contain above `TextDocumentEdit`s mixed with
	 * create, rename and delete file / folder operations.
	 *
	 * Whether a client supports versioned document edits is expressed via
	 * `workspace.workspaceEdit.documentChanges` client capability.
	 *
	 * If a client neither supports `documentChanges` nor
	 * `workspace.workspaceEdit.resourceOperations` then only plain `TextEdit`s
	 * using the `changes` property are supported.
	 */
	documentChanges?: (
		TextDocumentEdit[] |
		(TextDocumentEdit | CreateFile | RenameFile | DeleteFile)[]
	);

	/**
	 * A map of change annotations that can be referenced in
	 * `AnnotatedTextEdit`s or create, rename and delete file / folder
	 * operations.
	 *
	 * Whether clients honor this property depends on the client capability
	 * `workspace.changeAnnotationSupport`.
	 *
	 * @since 3.16.0
	 */
	changeAnnotations?: {
		[id: string /* ChangeAnnotationIdentifier */]: ChangeAnnotation;
	};
}
```

UziTech (changes requested):
LGTM just needs tests

UziTech:
@aminya  looks like linting is failing because of an issue with pnpm. Could you look into that?

aminya:
> @aminya looks like linting is failing because of an issue with pnpm. Could you look into that?

Yes, I will fix that. 
https://github.com/atom-community/atom-ide-hyperclick/issues/132#issuecomment-818310509

cdaguerre:
@UziTech I implemented handling of resource operation options and added tests.
Rollbacks of applied resource operations in case of error are not handled for now but I think it should ok for a first iteration.

test/adapters/apply-edit-adapter.test.ts
```diff
@@ -186,5 +189,235 @@ describe("ApplyEditAdapter", () => {
...
+      expect(fs.existsSync(directory)).to.equal(true)
+      expect(fs.existsSync(file1)).to.equal(true)
+      expect(fs.existsSync(file2)).to.equal(true)
+      const errorCalls = (atom as any).notifications.addError.getCalls()
+      expect(errorCalls.length).to.equal(1)
+      expect(errorCalls[0].args[1].detail).to.match(/Error during delete resource operation: (.*)/)
+    })
+
+
+    it("throws an error on delete operation if target doesnt exist", async () => {
```
UziTech, line 383: Should the delete operation be idempotent and fail silently if the file has already been deleted?

lib/adapters/apply-edit-adapter.ts
```diff
@@ -81,6 +90,86 @@ export default class ApplyEditAdapter {
     return { applied }
   }
 
+  private static async handleResourceOperation(edit: (CreateFile | RenameFile | DeleteFile)): Promise<void> {
+    if (DeleteFile.is(edit)) {
+      const path = Convert.uriToPath(edit.uri)
+      const exists = fs.existsSync(path)
```
UziTech, line 96: We should use async fs functions to prevent blocking the main thread.

UziTech (changes requested):

lib/adapters/apply-edit-adapter.ts
cdaguerre: Replaced by:
```typescript
const exists = await fs.promises.stat(path).then(() => true).catch(() => false)
```

test/adapters/apply-edit-adapter.test.ts
cdaguerre: I'm not sure.
LSP's `CreateFile` actually has an `ignoreIfNotExists` param.
What I implemented is:
- If `ignoreIfNotExists` is `undefined`, assume `false` and throw error
- If `ignoreIfNotExists` is `true`, return in case the file (or directory) doesn't exist

All ResourceOperations of such kind of params, notably `overwrite`, `ignoreIfExists` and `ignoreIfNotExists`.
I don't think that LSP gives any guidance on what should be considered the default value for such params, so I just assumed `undefined === false`, but I agree it might make sense to adapt with more sensible defaults...
WDYT?

lib/adapters/apply-edit-adapter.ts
UziTech: I think `fs.promises.access` would be better since we don't use the stat object.

aminya:
> > @aminya looks like linting is failing because of an issue with pnpm. Could you look into that?
> 
> Yes, I will fix that.
> [atom-community/atom-ide-hyperclick/issues/132#issuecomment-818310509](https://github.com/atom-community/atom-ide-hyperclick/issues/132#issuecomment-818310509)

#145 and #144 will fix the linting errors.

lib/adapters/apply-edit-adapter.ts
aminya: There is also the path-exits package that exactly does this.
https://www.npmjs.com/package/path-exists

UziTech: seems a little much to install a dependency for a one liner

test/adapters/apply-edit-adapter.test.ts
UziTech: @aminya what do you think? I am on the fence on this one. On the one hand it makes sense to have `undefined` == `false`. On the other hand I don't see any situation where someone would want an error when deleting a file if the file already doesn't exist.

lib/adapters/apply-edit-adapter.ts
cdaguerre: I agree with @UziTech 
I already find it annoying having to require `rimraf`;)

cdaguerre:
I changed the code to use `fs.stats` and fixed linting/format errors.
Build is green! 🎉

lib/adapters/apply-edit-adapter.ts
```diff
@@ -1,7 +1,19 @@
...
+  ApplyWorkspaceEditResponse,
+  WorkspaceEdit,
+  TextDocumentEdit,
+  CreateFile,
+  RenameFile,
+  DeleteFile,
+  DocumentUri,
+} from "../languageclient"
 import { TextBuffer, TextEditor } from "atom"
+import * as fs from "fs"
```
aminya, line 15: This cleans up deep calls to `fs.promises.*`
```suggestion
import { promises as fsp, Stats } from "fs"
```

lib/adapters/apply-edit-adapter.ts
```diff
@@ -30,36 +42,35 @@ export default class ApplyEditAdapter {
...
+            throw Error(`Error during ${edit.kind} resource operation: ${err.message}`)
+          })
+        }
+        const path = Convert.uriToPath(edit.textDocument.uri)
+        const editor = (await atom.workspace.open(path, {
+          searchAllPanes: true,
+          // Open new editors in the background.
+          activatePane: false,
+          activateItem: false,
+        })) as TextEditor
```
aminya, line 67: Any chance that this will return anything else?

test/adapters/apply-edit-adapter.test.ts
cdaguerre: I actually agree that a delete operation should not throw an error if the file doesn't exist.
For all other params/operation the sensible default would be `false` like : 
- A `rename` operation should not overwrite another file and fail by default if the target exists (defaults: `overwrite = false` , `ignoreIfExists = false`
- A `create` operation should not overwrite another file and fail by default if the target exists (defaults: `overwrite = false` , `ignoreIfExists = false`
- A `delete` operation should not fail by default if the target doesn't exist (defaults: `ignoreIfNotExists = true` **<== our case here**

So in case of a delete operation:
```typescript
const ignoreIfNotExists = edit.options?.ignoreIfNotExists || true
```
Agreed?

lib/adapters/apply-edit-adapter.ts
cdaguerre: Not really I guess, but here's the interface
```typescript
open(uri: string, options?: WorkspaceOpenOptions): Promise<object>;
```

lib/adapters/apply-edit-adapter.ts
```diff
@@ -78,6 +89,90 @@ export default class ApplyEditAdapter {
...
+
+      if (exists && ignoreIfExists && !overwrite) {
+        return
+      }
+
+      return fsp.writeFile(path, "")
+    }
+  }
+
+  private static normalize(workspaceEdit: WorkspaceEdit): void {
```
aminya, line 158: I am not a fan of using static methods. Would you be able to make these free functions? If public method, you can directly export them.

lib/adapters/apply-edit-adapter.ts
```diff
@@ -92,8 +92,8 @@ export default class ApplyEditAdapter {
   private static async handleResourceOperation(edit: CreateFile | RenameFile | DeleteFile): Promise<void> {
     if (DeleteFile.is(edit)) {
       const path = Convert.uriToPath(edit.uri)
-      const stats: boolean | fs.Stats = await fs.promises.lstat(path).catch(() => false)
-      const ignoreIfNotExists = edit.options?.ignoreIfNotExists
+      const stats: boolean | Stats = await fsp.lstat(path).catch(() => false)
+      const ignoreIfNotExists = edit.options?.ignoreIfNotExists || true
```
aminya, line 96: This newly added `true` seems to fail the test. We should edit one of them.

lib/adapters/apply-edit-adapter.ts
cdaguerre: What do you mean by free? Could you give an example?

lib/adapters/apply-edit-adapter.ts
aminya: Thank you. I might update the `@types/atom` later to reflect the edge cases.

lib/adapters/apply-edit-adapter.ts
aminya: For example, for the private methods, you can simply move the code to the bottom of the file.
```ts
function normalize(workspaceEdit: WorkspaceEdit): void {
 // code
}
```

aminya: If the method is already public, you can keep it for now. We have started to avoid using classes as a namespace, especially for new code.

cdaguerre: Like this ? https://github.com/atom-community/atom-languageclient/pull/140/commits/a699bd57ac50748a2c490de2bacc58f080423115

aminya: Exactly! Thanks!

aminya (approved):
Looks good to me. Thanks for your contribution!

test/adapters/apply-edit-adapter.test.ts
cdaguerre: @UziTech This is what I did: https://github.com/atom-community/atom-languageclient/pull/140/files#diff-9011b4bf132f9c2bff2dfede53c468cea4b1710ce6937a8948a733ca1f109c52R99
Basically the default behavior is: don't throw an error when deleting a file that doesn't exist, except if the server requested `ignoreIfNotExists: false`.

UziTech (approved):

PR #141, opened 2021-04-12
feat: update rename functionality

cdaguerre:
Hi,
To my understanding, the current rename implementation doesn't work anymore. 
I'm even wondering if it worked at some point because of the comment that @lexi-lambda made in the PR description: https://github.com/atom/atom-languageclient/pull/270

Anyways, it seemed to rely on the archived atom-ide-ui...

So here's a draft PR to suggest two different implementations:
- [**Intentions**](https://atom.io/packages/intentions) (https://github.com/cdaguerre/atom-languageclient/commit/35f957e0ce60c61ef0d4d405a02b8948058f048b)
- **Commands** and context menu (https://github.com/cdaguerre/atom-languageclient/commit/0516e36b087683b522d7166fae8edea89407e7c8)

Both approaches have advantages and drawbacks, eg. 

**Commands and context menu**
- You can't conditionally disable context menu items asynchronously: the refactor > rename item is always available and will trigger an error when no rename is possible at the given position
- I found no way to pass the mousevent click position to the command, so the command uses the text editor's buffer position which we set manually on context menu right click.

**Intentions**
- Feels less "native" 
- No right click possible, requires a keybinding
- Rename action is only suggested when possible (using LSP prepare functionality)


lib/views/dialog.ts
```diff
@@ -0,0 +1,37 @@
+import { TextEditor } from "atom"
+
+export default class Dialog {
```
UziTech, line 3: I actually created a dialog view for some other atom packages https://github.com/UziTech/atom-modal-views. Do we want to use that instead of creating our own?

/cc @aminya

aminya: > I actually created a dialog view for some other atom packages UziTech/atom-modal-views. Do we want to use that instead of creating our own?

That sounds like a good plan! I love reusing code.

aminya:
The last time I checked the intentions package was quite buggy. For example, it disabled selecting the text, or the underlines were longer than what they should be. I thought @steelbrain had stopped supporting it.
https://github.com/steelbrain/intentions/issues/76

aminya:
I got access to the intentions package. We should fix these issues before merging this PR.
https://github.com/steelbrain/intentions/issues/76
https://github.com/steelbrain/intentions/issues/74
https://github.com/steelbrain/intentions/issues/77

ivan94fi:
Hi, any updates on this functionality? Will this be merged in the near future?

Thank you all from atom-community for your work!

PR #146, opened 2021-04-16
fix: add `LocationLink` support for definition-adapter

ayame113:
According to https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_definition, the `textDocument/definition` request returns not only `Location[]` but also `LocationLink[]`. May I add support for this?

This is my first experience of TypeScript, so please let me know if I make any mistakes.

### DIFF
--- a/lib/adapters/definition-adapter.ts
+++ b/lib/adapters/definition-adapter.ts
@@ -107,6 +107,6 @@ export default class DefinitionAdapter {
     }))
   }
   private static isLocationLinkArray(value: any): value is LocationLink[] {
-    return Array.isArray(value) && LocationLink.is(value[0])
+    return Array.isArray(value) && value.every((v) => LocationLink.is(v))
   }
 }
--- a/lib/adapters/definition-adapter.ts
+++ b/lib/adapters/definition-adapter.ts
@@ -72,8 +72,10 @@ export default class DefinitionAdapter {
     locationResult: Location | Location[] | LocationLink[] | null
   ): Location[] | LocationLink[] | null {
     if (locationResult == null) {
+      // TODO use ===
       return null
     }
+    // TODO `d.targetRange.start` never becomes `null` according to the types
     if (isLocationLinkArray(locationResult)) {
       return locationResult.filter((d) => d.targetRange.start != null)
     }
--- a/lib/adapters/definition-adapter.ts
+++ b/lib/adapters/definition-adapter.ts
@@ -74,7 +74,7 @@ export default class DefinitionAdapter {
     if (locationResult == null) {
       return null
     }
-    if (this.isLocationLinkArray(locationResult)) {
+    if (isLocationLinkArray(locationResult)) {
       return locationResult.filter((d) => d.targetRange.start != null)
     }
     return (Array.isArray(locationResult) ? locationResult : [locationResult]).filter((d) => d.range.start != null)
@@ -91,7 +91,7 @@ export default class DefinitionAdapter {
     locations: Location[] | LocationLink[],
     languageName: string
   ): atomIde.Definition[] {
-    if (this.isLocationLinkArray(locations)) {
+    if (isLocationLinkArray(locations)) {
       return locations.map((d) => ({
         path: Convert.uriToPath(d.targetUri),
         position: Convert.positionToPoint(d.targetRange.start),
@@ -106,7 +106,8 @@ export default class DefinitionAdapter {
       language: languageName,
     }))
   }
-  private static isLocationLinkArray(value: any): value is LocationLink[] {
-    return Array.isArray(value) && value.every((v) => LocationLink.is(v))
-  }
+}
+
+function isLocationLinkArray(value: any): value is LocationLink[] {
+  return Array.isArray(value) && LocationLink.is(value[0])
 }
--- a/lib/adapters/definition-adapter.ts
+++ b/lib/adapters/definition-adapter.ts
@@ -74,9 +74,10 @@ export default class DefinitionAdapter {
     if (locationResult == null) {
       return null
     }
-    return (Array.isArray(locationResult) ? (locationResult as any[]) : [locationResult]).filter(
-      (d: Location | LocationLink) => ("range" in d ? d.range : d.targetRange).start != null
-    )
+    if (this.isLocationLinkArray(locationResult)) {
+      return locationResult.filter((d) => d.targetRange.start != null)
+    }
+    return (Array.isArray(locationResult) ? locationResult : [locationResult]).filter((d) => d.range.start != null)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lib/adapters/definition-adapter.ts
```diff
@@ -68,11 +68,15 @@ export default class DefinitionAdapter {
    * @param locationResult Either a single {Location} object or an {Array} of {Locations}.
    * @returns An {Array} of {Location}s or {null} if the locationResult was null.
    */
-  public static normalizeLocations(locationResult: Location | Location[]): Location[] | null {
+  public static normalizeLocations(
+    locationResult: Location | (Location | LocationLink)[]
+  ): (Location | LocationLink)[] | null {
```
aminya, line 73: Based on the docs, this is either `Location` or `LocationLink`. It is not a mixed array.
```suggestion
    locationResult: Location | Location[] | LocationLink[]
  ): Location[] | LocationLink[] | null {
```

lib/adapters/definition-adapter.ts
```diff
@@ -82,11 +86,14 @@ export default class DefinitionAdapter {
    * @param languageName The name of the language these objects are written in.
    * @returns An {Array} of {Definition}s that represented the converted {Location}s.
    */
-  public static convertLocationsToDefinitions(locations: Location[], languageName: string): atomIde.Definition[] {
+  public static convertLocationsToDefinitions(
+    locations: (Location | LocationLink)[],
```
aminya, line 90:
```suggestion
    locations: Location[] | LocationLink[],
```

lib/adapters/definition-adapter.ts
ayame113: Thank you for review.
When I make this change, I get the following error:
```ts
lib/adapters/definition-adapter.ts:77:80 - error TS2349: This expression is not callable.
  Each member of the union type '{ <S extends Location>(predicate: (value: Location, index: number, array: Location[]) => value is S, thisArg?: any): S[]; (predicate: (value: Location, index: number, array: Location[]) => unknown, thisArg?: any): Location[]; } | { ...; }' has signatures, but none of those signatures are compatible with each other.

77     return (Array.isArray(locationResult) ? locationResult : [locationResult]).filter(
                                                                                  ~~~~~~

lib/adapters/definition-adapter.ts:78:8 - error TS7006: Parameter 'd' implicitly has an 'any' type.

78       (d) => ("range" in d ? d.range : d.targetRange).start != null
          ~
```

For example, is it okay to add  `as any []`  as shown below?

```ts
  public static normalizeLocations(
    locationResult: Location | Location[] | LocationLink[] | null
  ): Location[] | LocationLink[] | null {
    if (locationResult == null) {
      return null
    }
    return (Array.isArray(locationResult) ? locationResult as any[] : [locationResult]).filter(
                                                           ^^^^^^^^^
      (d) => ("range" in d ? d.range : d.targetRange).start != null
    )
  }
```

UziTech: I think you should add a type to `d`. Something like:

```js
(d: Location | LocationLink) => ("range" in d ? d.range : d.targetRange).start != null
```

ayame113: Fixed in the way you said.

lib/adapters/definition-adapter.ts
ayame113: I fixed it in the same way as above.

lib/adapters/definition-adapter.ts
```diff
@@ -68,11 +68,15 @@ export default class DefinitionAdapter {
...
    */
-  public static normalizeLocations(locationResult: Location | Location[]): Location[] | null {
+  public static normalizeLocations(
+    locationResult: Location | Location[] | LocationLink[] | null
+  ): Location[] | LocationLink[] | null {
     if (locationResult == null) {
       return null
     }
-    return (Array.isArray(locationResult) ? locationResult : [locationResult]).filter((d) => d.range.start != null)
+    return (Array.isArray(locationResult) ? (locationResult as any[]) : [locationResult]).filter(
```
aminya, line 77: This cast to `any` is very suspicious. Casting shouldn't be used in normal cases.

lib/adapters/definition-adapter.ts
```diff
@@ -68,11 +68,15 @@ export default class DefinitionAdapter {
...
+      (d: Location | LocationLink) => ("range" in d ? d.range : d.targetRange).start != null
```
aminya, line 78: Because the array is not a mixed array, this check should only be performed for the first element of the array. There is no need to check it for every element.

lib/adapters/definition-adapter.ts
```diff
@@ -82,11 +86,14 @@ export default class DefinitionAdapter {
...
-    return locations.map((d) => ({
-      path: Convert.uriToPath(d.uri),
-      position: Convert.positionToPoint(d.range.start),
-      range: Range.fromObject(Convert.lsRangeToAtomRange(d.range)),
+  public static convertLocationsToDefinitions(
+    locations: Location[] | LocationLink[],
+    languageName: string
+  ): atomIde.Definition[] {
+    return (locations as any[]).map((d: Location | LocationLink) => ({
+      path: Convert.uriToPath("uri" in d ? d.uri : d.targetUri),
```
aminya, line 94: The same comment as above.

aminya (changes requested):

lib/adapters/definition-adapter.ts
ayame113: Instead of not using the casting, I added a type guard function.

lib/adapters/definition-adapter.ts
```diff
@@ -82,12 +87,26 @@ export default class DefinitionAdapter {
...
+    }
     return locations.map((d) => ({
       path: Convert.uriToPath(d.uri),
       position: Convert.positionToPoint(d.range.start),
       range: Range.fromObject(Convert.lsRangeToAtomRange(d.range)),
       language: languageName,
     }))
   }
+  private static isLocationLinkArray(value: any): value is LocationLink[] {
+    return Array.isArray(value) && value.every((v) => LocationLink.is(v))
```
aminya, line 110: As mentioned, there is no need to check every single element.
```suggestion
    return Array.isArray(value) && LocationLink.is(value?.[0])
```

aminya (changes requested):

lib/adapters/definition-adapter.ts
```diff
@@ -82,12 +87,26 @@ export default class DefinitionAdapter {
...
+      }))
+    }
     return locations.map((d) => ({
       path: Convert.uriToPath(d.uri),
       position: Convert.positionToPoint(d.range.start),
       range: Range.fromObject(Convert.lsRangeToAtomRange(d.range)),
       language: languageName,
     }))
   }
+  private static isLocationLinkArray(value: any): value is LocationLink[] {
```
aminya, line 109: This should be a free function instead of a class method. 
```ts
function isLocationLinkArray(value: any): value is LocationLink[] {
```

aminya (changes requested):
Thanks for your contribution! 🚀  I added two more comments about the implementation.

lib/adapters/definition-adapter.ts
ayame113: Fixed it

lib/adapters/definition-adapter.ts
ayame113: Fixed it too.

aminya (approved):
This looks good to me! 

I have just one final question. Shouldn't we change anything in the capabilities list?

https://github.com/atom-community/atom-languageclient/blob/9abd5b6c813189bc126a6636ffcd2a42f8c6e5aa/lib/auto-languageclient.ts#L117

ayame113:
In my opinion, `textDocument/definition` is already enabled, so I don't think it needs to be changed.

https://github.com/atom-community/atom-languageclient/blob/9abd5b6c813189bc126a6636ffcd2a42f8c6e5aa/lib/auto-languageclient.ts#L209-L211

ayame113:
thank you!👍

aminya:
Thank **you** for your contribution!

PR #148, opened 2021-04-20
fix: set the correct fallback for Snippet completion

ayame113:
Currently, Snippets without `insertText` are not completed (empty strings are inserted).
According to the lsp `CompletionItem` spec, Implementation should fall back to `label`.
This pull request fixes it.

>https://microsoft.github.io/language-server-protocol/specification#textDocument_completion
>A string that should be inserted into a document when selecting this completion. When `falsy` the label is used as the insert text for this item.


UziTech (changes requested):
Thanks! Could you add a test?

ayame113:
Thank you for review.
Added a test. However, I'm new to writing tests so I don't know if this is correct ...

UziTech (approved):

UziTech:
Looks great! :100: Thanks for the PR!

ayame113:
Thank you!

PR #149, opened 2021-04-21
fix: update deps

UziTech:


lib/adapters/autocomplete-adapter.ts
```diff
@@ -133,7 +133,7 @@ export default class AutocompleteAdapter {
     }
 
     const filtered = !(request.prefix === "" || (triggerChar !== "" && triggerOnly))
-    return filtered ? filter(suggestions, request.prefix, { key: "filterText" }) : suggestions
+    return filtered ? filter<any>(suggestions, request.prefix, { key: "filterText" }) : suggestions
```
UziTech, line 136: @aminya I don't know if zadeh types need to be changed, but putting `Suggestion` here makes typescript fail. Not the end of the world but I would rather not have an `any` here.

aminya:
I think you can ignore zadeh for now. The version 3 is becoming ready.

aminya:
OK. I updated `Zadeh` to `3.0.0-beta.2` which uses `ObjectArrayFilterer`. We can highly improve the performance by only setting the suggestion when the cache is not hit.

lib/adapters/autocomplete-adapter.ts
```diff
@@ -133,7 +133,17 @@ export default class AutocompleteAdapter {
     }
 
     const filtered = !(request.prefix === "" || (triggerChar !== "" && triggerOnly))
-    return filtered ? filter(suggestions, request.prefix, { key: "filterText" }) : suggestions
+    if (filtered) {
+      // filter the suggestions who have `filterText` property
+      const validSuggestions = suggestions.filter((sgs) => typeof sgs.filterText === "string") as Suggestion[] &
```
aminya, line 138: Probably, I can move this check to Zadeh itself.

UziTech: Ya I would think zadeh should treat objects that don't have the `key` the same as an object having an empty string as the key.

aminya: But now that I think this shouldn't be done. For example, `StringArrayFilterer` expects an array of strings, and no check is done to ensure that every element is a string. So, the same thing should be valid for `ObjectArrayFilterer` for consistency and also performance. I think it is the responsibility of the user to conform to the type-safety of Zadeh, not the reverse.

PR #150, opened 2021-04-24
Added null check to some adapters and changed type definition

ayame113:
Some LSP requests lack null checking and are causing errors, so I fixed them.

image

I made the following changes:
1. Added null check (https://github.com/atom-community/atom-languageclient/commit/e20bba5a4fe06b4f1a24f2063b1a60fcd5da1514)
2. Fixed the type definition (https://github.com/atom-community/atom-languageclient/commit/85cc505a5c9999e4da044996aa49e436d32d017a)
3. Use the vscode-languageserver-protocol package for type definition (https://github.com/atom-community/atom-languageclient/compare/85cc505..d665159)

If you don't need the third change, reverting it will still work.

### DIFF
--- a/lib/adapters/code-action-adapter.ts
+++ b/lib/adapters/code-action-adapter.ts
@@ -39,7 +39,7 @@ export default class CodeActionAdapter {
     editor: TextEditor,
     range: Range,
     diagnostics: atomIde.Diagnostic[]
-  ): Promise<atomIde.CodeAction[] | null> {
+  ): Promise<atomIde.CodeAction[]> {
     if (linterAdapter == null) {
       return []
     }
@@ -48,7 +48,7 @@ export default class CodeActionAdapter {
     const params = CodeActionAdapter.createCodeActionParams(linterAdapter, editor, range, diagnostics)
     const actions = await connection.codeAction(params)
     if (!actions) {
-      return null
+      return []
     }
     return actions.map((action) => CodeActionAdapter.createCodeAction(action, connection))
   }
--- a/lib/adapters/code-action-adapter.ts
+++ b/lib/adapters/code-action-adapter.ts
@@ -39,7 +39,7 @@ export default class CodeActionAdapter {
     editor: TextEditor,
     range: Range,
     diagnostics: atomIde.Diagnostic[]
-  ): Promise<atomIde.CodeAction[]> {
+  ): Promise<atomIde.CodeAction[] | null> {
     if (linterAdapter == null) {
       return []
     }
@@ -47,7 +47,10 @@ export default class CodeActionAdapter {
 
     const params = CodeActionAdapter.createCodeActionParams(linterAdapter, editor, range, diagnostics)
     const actions = await connection.codeAction(params)
-    return (actions || []).map((action) => CodeActionAdapter.createCodeAction(action, connection))
+    if (!actions) {
+      return null
+    }
+    return actions.map((action) => CodeActionAdapter.createCodeAction(action, connection))
   }
 
   private static createCodeAction(
--- a/lib/adapters/code-action-adapter.ts
+++ b/lib/adapters/code-action-adapter.ts
@@ -47,7 +47,7 @@ export default class CodeActionAdapter {
 
     const params = CodeActionAdapter.createCodeActionParams(linterAdapter, editor, range, diagnostics)
     const actions = await connection.codeAction(params)
-    return actions.map((action) => CodeActionAdapter.createCodeAction(action, connection))
+    return (actions||[]).map((action) => CodeActionAdapter.createCodeAction(action, connection))
   }
 
   private static createCodeAction(
--- a/lib/adapters/code-action-adapter.ts
+++ b/lib/adapters/code-action-adapter.ts
@@ -47,7 +47,7 @@ export default class CodeActionAdapter {
 
     const params = CodeActionAdapter.createCodeActionParams(linterAdapter, editor, range, diagnostics)
     const actions = await connection.codeAction(params)
-    return (actions||[]).map((action) => CodeActionAdapter.createCodeAction(action, connection))
+    return (actions || []).map((action) => CodeActionAdapter.createCodeAction(action, connection))
   }
 
   private static createCodeAction(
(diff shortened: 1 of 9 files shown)


UziTech:
I think we were discussing this in https://github.com/atom-community/atom-languageclient/pull/132#discussion_r595580845

I'm not sure if it is better to through an error if the lsp returns something invalid or fail silently like this PR. Is null or undefined a valid return for `.codeAction`?

aminya:
> I think we were discussing this in [#132 (comment)](https://github.com/atom-community/atom-languageclient/pull/132#discussion_r595580845)

This is different though.

UziTech:
> This is different though.

How? Why are these checks necessary if `null` or `undefined` are not correct outputs of `.codeAction` anyway?

aminya:
I meant the changes under `lib/languageclient.ts` which seem reasonable as they are using the original types.

UziTech:
> I meant the changes under `lib/languageclient.ts` which seem reasonable as they are using the original types.

Ya I think those changes are good. I'm just not sure about null checks.

ayame113:
(some comment)
- Currently, the [`vscode-languageserver-protocol` module](https://www.npmjs.com/package/vscode-languageserver-protocol) is used for type definition. In this PR, I also want to use it in the return value of the language-client method.
 - Since the return type is not exported directly from the module, I am using the generics of the function.
   - I used it as a reference for implementation
https://github.com/microsoft/vscode-languageserver-node/blob/d58c00bbf8837b9fd0144924db5e7b1c543d839e/client/src/common/client.ts#L83-L86
 - This will cause a type error if the wrong return type.
    image
- Removed null from the return value of the following request according to the type definition
  - completionItem / resolve
  - codeLens / resolve
 - Added null
   - textDocument/references
   - textDocument/documentHighlight
   - textDocument/documentSymbol
   - workspace/symbol
   - textDocument/codeAction
   - textDocument/codeLens
   - textDocument/documentLink
   - textDocument/formatting
   - textDocument/rangeFormatting
   - textDocument/onTypeFormatting
   - textDocument/rename
 - I think this implementation fits the spec, but you can leave null in the return value if you want to deal with invalid null checks outside the spec.

lib/adapters/code-action-adapter.ts
```diff
@@ -47,7 +47,7 @@ export default class CodeActionAdapter {
 
     const params = CodeActionAdapter.createCodeActionParams(linterAdapter, editor, range, diagnostics)
     const actions = await connection.codeAction(params)
-    return actions.map((action) => CodeActionAdapter.createCodeAction(action, connection))
+    return (actions || []).map((action) => CodeActionAdapter.createCodeAction(action, connection))
```
UziTech, line 50: According to [`codeAction`](https://github.com/atom-community/atom-languageclient/blob/ffbe675bf8a2dc90171a5599dd2dcf88a4398f0f/lib/languageclient.ts#L403) it returns `Promise<Array<lsp.Command | lsp.CodeAction>>` so a promise resolving to `null` or `undefined` is an invalid return type and should probably throw an error instead of being handled the same as an empty array. Or the return type should be updated to include `null` and `undefined` as valid returns.

UziTech: I tend to prefer throwing a valid error instead of failing silently like this. @aminya This is what I was talking about in https://github.com/atom-community/atom-languageclient/pull/132#discussion_r595452569 when I said "types do not catch runtime errors".

lib/languageclient.ts
```diff
@@ -421,8 +421,8 @@ export class LanguageClientConnection extends EventEmitter {
    * @param params The {CodeLens} identifying the code lens to be resolved with full detail.
    * @returns A {Promise} containing the {CodeLens} fully resolved.
    */
-  public codeLensResolve(params: lsp.CodeLens): Promise<lsp.CodeLens | null> {
-    return this._sendRequest("codeLens/resolve", params)
+  public codeLensResolve(params: lsp.CodeLens): Promise<lsp.CodeLens> {
+    return this._sendRequest(lsp.CodeLensResolveRequest.type, params)
```
UziTech, line 425: Would this be a breaking change since we are changing a public API?

lib/adapters/code-action-adapter.ts
ayame113: Thank you for your review!

The spec seems to say that codeAction returns `(Command | CodeAction) [] | null`. Is it not necessary to raise an error if the language server returns valid null?
> https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction
> Request:
> - method: ‘textDocument/codeAction’
> 
> Response:
> - result: (Command | CodeAction)[] | null

By the way, such a change is also an option.
```diff
 public static async getCodeActions(
   ...
-): Promise<atomIde.CodeAction[]> {
+): Promise<atomIde.CodeAction[] | null> {
   if (linterAdapter == null) {
     return []
   }
   assert(serverCapabilities.codeActionProvider, "Must have the textDocument/codeAction capability")
   const params = CodeActionAdapter.createCodeActionParams(linterAdapter, editor, range, diagnostics)
   const actions = await connection.codeAction(params)
-  return (actions || []).map((action) => CodeActionAdapter.createCodeAction(action, connection))
+  if (actions==null) {return null}
+  return actions.map((action) => CodeActionAdapter.createCodeAction(action, connection))
 }
```

lib/languageclient.ts
ayame113: The LanguageClient class has not been exported from main.ts, but it may still be a breaking change. I do not know well which.😰

lib/adapters/code-action-adapter.ts
UziTech: I like the early null check rather than still mapping over an empty array.

We should change the return type of [codeAction](https://github.com/atom-community/atom-languageclient/blob/ffbe675bf8a2dc90171a5599dd2dcf88a4398f0f/lib/languageclient.ts#L403) if null is a valid return according to the spec.

lib/languageclient.ts
UziTech: Maybe just a note in the changelogs

UziTech (approved):

aminya:
I'll review today or tomorrow

lib/adapters/code-action-adapter.ts
ayame113: At https://github.com/atom-community/atom-languageclient/pull/150/commits/e7c26b4f22dbf58ae17788b2a66fa04abeafc466, I changed it to return null instead of an empty array when `actions` value is null.

aminya (approved):

aminya (approved):

PR #153, opened 2021-05-06
feat: add support for workspace folders

aminya:
Adds support for WorkspaceFolders as described here:

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_workspaceFolders


lib/server-manager.ts
```diff
@@ -255,14 +255,50 @@ export class ServerManager {
...
-    const pathsSet = new Set(projectPaths.map(this.normalizePath))
-    const serversToStop = this._activeServers.filter((s) => !pathsSet.has(s.projectPath))
+    const pathsSet = projectPaths.map(this.normalizePath)
+    const pathsRemoved = this.getProjectPaths().filter((projectPath) => !pathsSet.includes(projectPath))
+
+    // send didChangeWorkspaceFolders
+    for (const activeServer of this._activeServers) {
+      activeServer.connection.didChangeWorkspaceFolders({
+        event: {
+          added: pathsSet.map(projectPathToWorkspaceFolder),
```
aminya, line 291: Not sure if this `pathsSet` is all the current project paths or only those that are added.

test/auto-languageclient.test.ts
```diff
@@ -1,24 +1,70 @@
...
+        })
+        it("returns null when a single file is open", async () => {
+          await atom.workspace.open(__filename)
+          const workspaceFolders = await serverManager.getWorkspaceFolders()
+          expect(workspaceFolders).to.be.null
+        })
+        it("gives the open workspace folders", async () => {
+          atom.project.addPath(__dirname)
+          serverManager.startServer(__dirname)
+          const workspaceFolders = await serverManager.getWorkspaceFolders()
```
aminya, line 51: Not sure why it still returns `null`.

lib/server-manager.ts
aminya: https://github.com/atom-community/atom-languageclient/pull/153/commits/4bce8629804ad6abe72493d2b66d94be62b2e4ee

test/auto-languageclient.test.ts
aminya: Oh, may bad. I have to await `startServer`.

typings/spawk/index.d.ts
```diff
@@ -0,0 +1,130 @@
+declare module "spawk" {
```
aminya, line 1: I made a pull request for this. We can remove it here once that one is merged
https://github.com/wraithgar/spawk/pull/10

lib/server-manager.ts
```diff
@@ -60,6 +60,7 @@ export class ServerManager {
         this._disposable.add(atom.project.onDidChangeFiles(this.projectFilesChanged.bind(this)))
       }
     }
+    this._isStarted = true
```
aminya, line 63: This was an irrelevant bug that I discovered here.

lib/server-manager.ts
```diff
@@ -252,17 +253,57 @@ export class ServerManager {
...
+
+    const previousPaths = this.getProjectPaths()
+    const pathsRemoved = previousPaths.filter((projectPath) => !pathsAll.includes(projectPath))
+    const pathsAdded = pathsAll.filter((projectPath) => !previousPaths.includes(projectPath))
+
+    // send didChangeWorkspaceFolders
+    for (const activeServer of this._activeServers) {
+      //@ts-ignore
+      console.log({spy: activeServer.connection.didChangeWorkspaceFolders.calls, pathsAdded, pathsRemoved})
+      activeServer.connection.didChangeWorkspaceFolders({
```
aminya, line 294: Not sure why it is not called in the `didChangeWorkspaceFolders` test

test/auto-languageclient.test.ts
```diff
@@ -1,25 +1,106 @@
...
+          server.connection.didChangeWorkspaceFolders = spy
+
+          atom.project.addPath(projectPath2)
+          console.log(spy.getCalls())
+
+          expect(spy.calledOnce).to.be.true
+          expect(spy.firstCall.args[0]).to.deep.equal({
+            added: [workspaceFolder2],
+            removed: []
+          })
```
UziTech, line 84: You forgot the `event` property.

```suggestion
          expect(spy.firstCall.args[0]).to.deep.equal({
            event: {
              added: [workspaceFolder2],
              removed: []
            }
          })
```

UziTech: also seems like maybe a race condition. If I run just this test it passes but running all tests makes this test fail.

aminya: Thanks. I had to tear down the tests inside afterEach.

aminya:
Atom crashes inside the CI on Ubuntu and MacOS after the tests are down. It might be related to this issue:
https://github.com/atom/atom/issues/22024

test/auto-languageclient.test.ts
```diff
@@ -1,25 +1,122 @@
...
+
+          atom.project.removePath(projectPath2)
+          expect(stub.calledTwice).to.be.true
+
+          expect(stub.secondCall.args[0]).to.deep.equal({
+            event: {
+              added: [],
+              removed: [workspaceFolder2],
+            },
+          })
```
aminya, line 100: This fails because for some reason `previousPaths` is already up to date! So removed is always empty.

UziTech: We should really switch the tests to atom-jasmine3-test-runner. I don't think the mocha test runner resets atom between tests.

UziTech: I'll work on a PR for that tonight

aminya: That would be great. However, this happens inside one test.

aminya: I don't think notifying about `removed` is actually needed because we already stop the servers if their project is removed. So, it seems safe to set this test as pending and merge this pull request.

aminya: OK. I fixed the issue for now by keeping a separate cache.

aminya (commented):
This is ready to go. We can test this with a real LSP to see how it works in practice.

aminya:
I tested it with `ide-d` and it seems that it works.

UziTech:
We should get the tests passing before we merge this otherwise it is going to be harder to know when tests for other PRs are passing. Even if we have to skip some tests.

aminya:
> We should get the tests passing before we merge this otherwise it is going to be harder to know when tests for other PRs are passing. Even if we have to skip some tests.

I made the tests working. I'm on Windows, so not sure why the CI fails on Ubuntu and MacOS.

test/auto-languageclient.test.ts
```diff
@@ -1,24 +1,128 @@
...
+          expect(spy.calls.mostRecent().args[0]).toEqual({
+            event: {
+              added: [],
+              removed: [workspaceFolder2],
+            },
+          })
+        })
+      })
+    })
+  })
```
aminya, line 110: These tests time out for some reason.

aminya: This commit fixes it, but not sure why I need to call it manually. 

https://github.com/atom-community/atom-languageclient/pull/153/commits/ea66c7fdea47b1e86bd65c80d60f105467598ec0

test/auto-languageclient.test.ts
```diff
@@ -1,24 +1,132 @@
...
+        client = new FakeAutoLanguageClient()
+        client.activate()
+
+        /* eslint-disable-next-line dot-notation */
+        serverManager = client["_serverManager"]
+      }
+
+      async function afterEachCallback() {
+        serverManager.stopListening()
+        await serverManager.stopAllServers()
```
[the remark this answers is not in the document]
aminya: It was working with Sinon. I think this is related to the way jasmine mocks things.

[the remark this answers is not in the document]
aminya: Does any of the Jasmine plugins modify `clearTimeout` and `setTimeout` functions? I had the exact same problem in [this PR](https://github.com/atom-community/atom-ide-outline/pull/118)

This code was working with Mocha, Sinon, Chai with not timing out.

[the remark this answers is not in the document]
UziTech: > Does any of the Jasmine plugins modify clearTimeout and setTimeout functions?

Not unless `jasmin.clock.install()` is called.

> Does Jasmine support async functions inside afterEach and beforeEach?

yes

> This code was not timing out with Mocha, Sinon, Chai

This was most likely because of a difference in sinon. `sinon.spy` calls the function by default and `spyOn` stubs the function by default.

test/helpers.ts
```diff
@@ -1,24 +1,37 @@
 import * as rpc from "vscode-jsonrpc"
 import { TextEditor } from "atom"
+import AutoLanguageClient from "../lib/auto-languageclient"
+import { LanguageClientConnection, Disposable, Event } from "../lib/languageclient"
+import { LanguageServerProcess } from "../lib/server-manager"
+import { spawn } from "spawk"
```
UziTech, line 6: I don't think we actually need spawk with jasmine.

```js
spyOn(require("child_process"), "spawn").and.callThrough()
```

That should be all we need.

aminya: But `spawn` will fail with this. We are not spawning a real language process.

test/helpers.ts
```diff
@@ -1,24 +1,37 @@
...
-    onNotification: jasmine.createSpy("onNotification"),
-    onProgress: jasmine.createSpy("onProgress"),
-    dispose: jasmine.createSpy("dispose"),
-    sendRequest: jasmine.createSpy("sendRequest"),
-    sendNotification: jasmine.createSpy("sendNotification"),
-    sendProgress: jasmine.createSpy("sendProgress"),
-    trace: jasmine.createSpy("trace"),
-    inspect: jasmine.createSpy("inspect"),
-    end: jasmine.createSpy("end"),
+    listen: jasmine.createSpy("listen").and.returnValue(void 0),
```
UziTech, line 19: why return `void 0` it should already return `undefined` by default.

test/helpers.ts
UziTech: Ya I guess spawk handles a lot of the mocking process.

test/helpers.ts
```diff
@@ -31,3 +44,42 @@ export function createFakeEditor(path?: string): TextEditor {
...
+/** A function that sleeps for a preiod of `ms` and then resolves the given value */
+function sleep<T>(resolveValue: T, ms: number) {
+  return new Promise<T>((resolve) => {
+    setTimeout(() => resolve(resolveValue), ms)
+  })
+}
+
+export function createFakeLanguageServerProcess(): LanguageServerProcess {
+  spawn("lsp", [], { stdio: "inherit", shell: true })
+    .exit(() => sleep(0, 500))
```
UziTech, line 57: It doesn't look like `exit` actually awaits this promise.
https://github.com/wraithgar/spawk/blob/main/lib/interceptor.js#L115

```suggestion
    .exit(0)
```

UziTech (changes requested):

test/helpers.ts
aminya: No, it does
https://github.com/wraithgar/spawk/blob/1baa72ad31d8cf3e439c0882f91bcfe8d3ba44f6/lib/interceptor.js#L211

test/helpers.ts
aminya: That's for clarity as it explicitly shows the intention of the spy.

lib/server-manager.ts
```diff
@@ -111,8 +113,8 @@ export class ServerManager {
     }
   }
 
-  public getActiveServers(): ActiveServer[] {
-    return this._activeServers.slice()
+  public getActiveServers(): Readonly<ActiveServer[]> {
+    return this._activeServers
```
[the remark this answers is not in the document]
aminya: I don't think doing slice helps here in any way here. If someone writes bad code which modifies the server, it is their mistake, and the code is still malfunctioning. Throwing a `slice` on an inherently unsafe architecture doesn't change anything. In interpreted languages like JavaScript, we should rely on the types, docs, etc since the runtime doesn't provide anything.

UziTech (approved):

aminya:
Please do not squash the pull requests. I need to disable squash for this organization altogether so people don't do it by mistake

UziTech:
Why not?

aminya:
We should not sacrifice the version control for the sake of semantic release change log generation. If the issue is that, we should stop using automatically generated release notes.

Related to this:
https://github.com/atom/atom/issues/22081

UziTech:
> but because it is squashed there is no way to see what issue comes from which change

Not true, this PR will still exist with the commits. And it will be easier to find since it will be in the release notes.

> people are used to assessing the life of a project based on the commits pushed to the master

That seems like an error on their part. Like judging software by number of lines of code or [“measuring aircraft building progress by weight”](https://twitter.com/codewisdom/status/837749167049031681?lang=en)

--- 

By the way, both of those arguments seem contrary to your arguments for combining dependency updates.

Issue #156, opened 2021-05-13
Merge alexheretic/atom-languageclient/tree/build to support ide-rust


aminya:
We need to merge the changes by @alexheretic from this branch, so ide-rust can switch to our fork
~https://github.com/alexheretic/atom-languageclient/tree/build~
https://github.com/atom-community/atom-languageclient/tree/ide-rust-build-no-rebuild (cleaned version)

- [x] [definitions for out-of-project files](https://github.com/atom-community/atom-languageclient/commit/6188c19888f285c42986764a569567a2df47e7cf) -> #160 
- [x] [Add shutdownServersGracefully() config function](https://github.com/atom-community/atom-languageclient/commit/11793871a00a8329bb53e1270d931768f70cd7d4) -> #161
- [x] [Fix completion deleting the triggerChar](https://github.com/atom-community/atom-languageclient/commit/0dcdd696b865ad72f0efcdb2a628b33d668dbee2) - already fixed in #114 
- [x] [Implement completion additionalTextEdits](https://github.com/atom-community/atom-languageclient/commit/1df5bd582e4109002771c42ed24de2197a6b0488) -> #162
- [x] [Support non-boolean *Provider server capabilities](https://github.com/atom-community/atom-languageclient/commit/5319279d68cf8837bc46cdf62975886c6b79052b) -> Already fixed
- [x] [send codeActionLiteralSupport](https://github.com/atom-community/atom-languageclient/commit/d5004ed2a159d165d2bf438161e9717f69b1c6c2) -> #167
- [x] [Support custom filterCodeActions & onApplyCodeActions](https://github.com/atom-community/atom-languageclient/commit/3ea785781ac299e4a0f5d4500bc5aedcb15399bf) -> #167

Allows fixing https://github.com/rust-lang/atom-ide-rust/issues/175

aminya:
I created a branch here:
https://github.com/atom-community/atom-languageclient/tree/ide-rust-build

PR #160, opened 2021-06-10
feat: support out of the project files

aminya:
- Moves `determineProjectPath` to `AutoLanguageClient` and passes it down to `ServerManager` in its constructor. (supercedes #11)
- Support for out-of-project files (modified from #156)
- Populate additional paths based on the definitions for such files (modified from #156)

### DIFF
--- a/test/auto-languageclient.test.ts
+++ b/test/auto-languageclient.test.ts
@@ -36,6 +36,11 @@ function setupServerManager(client = setupClient()) {
 describe("AutoLanguageClient", () => {
   describe("determineProjectPath", () => {
     it("returns the project path for an internal or an external file in the project", async () => {
+      if (process.platform === "darwin") {
+        // there is nothing OS specific about the code. It just hits the limits that MacOS can handle in this test
+        pending("skipped on MacOS")
+        return
+      }
       const client = setupClient()
       const serverManager = setupServerManager(client)
 
--- a/test/auto-languageclient.test.ts
+++ b/test/auto-languageclient.test.ts
@@ -35,43 +35,31 @@ function setupServerManager(client = setupClient()) {
 
 describe("AutoLanguageClient", () => {
   describe("determineProjectPath", () => {
-    it("returns null when a single file is open", async () => {
-      const client = setupClient()
-      const textEditor = (await atom.workspace.open(__filename)) as TextEditor
-      /* eslint-disable-next-line dot-notation */
-      const projectPath = client["determineProjectPath"](textEditor)
-      expect(projectPath).toBeNull()
-    })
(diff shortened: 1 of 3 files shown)


UziTech:
We should add tests before this is merged.

aminya:
Ready to go

test/auto-languageclient.test.ts
```diff
@@ -12,20 +19,91 @@ function mockEditor(uri: string, scopeName: string): any {
...
+
+      const textEditor = (await atom.workspace.open(__filename)) as TextEditor
+      /* eslint-disable-next-line dot-notation */
+      expect(client["determineProjectPath"](textEditor)).toBe(normalizePath(projectPath))
+    })
+    it("returns the project path for an external file if it is in additional paths", async () => {
+      // macos has issues with handling too much resources
+      if (process.platform === "darwin") {
+        return
+      }
```
UziTech, line 67: We should mark them as pending so they are easier to track.

```suggestion
      if (process.platform === "darwin") {
        pending("skipped on macos")
        return
      }
```

UziTech (changes requested):

test/auto-languageclient.test.ts
```diff
@@ -12,20 +19,91 @@ function mockEditor(uri: string, scopeName: string): any {
...
+  describe("determineProjectPath", () => {
+    it("returns null when a single file is open", async () => {
+      const client = setupClient()
+      const textEditor = (await atom.workspace.open(__filename)) as TextEditor
+      /* eslint-disable-next-line dot-notation */
+      const projectPath = client["determineProjectPath"](textEditor)
+      expect(projectPath).toBeNull()
+    })
+    it("returns the project path when a file of that project is open", async () => {
+      // macos has issues with handling too much resources
```
UziTech, line 46: is this just an issue on GitHub actions or would this crash Atom if run on a Mac?

test/auto-languageclient.test.ts
aminya: Fixed in 
https://github.com/atom-community/atom-languageclient/pull/160/commits/0f29b94dcf5b0ed49b134b48e49980f7916d8581

test/auto-languageclient.test.ts
aminya: Because the number of tests is high, MacOS can't run all of them. It works fine if each of the tests runs in a single process. There is nothing OS-specific in the code, so this is a limitation of the operating system.

test/auto-languageclient.test.ts
aminya: It seems `pending` cannot be added to describe. So I had to add it only to the `it`s of this PR.

UziTech (approved):

PR #161, opened 2021-06-12
feat: allow disabling shutdownGracefully

aminya:
Requires #160 (only the last commit is related to this PR). It was possible to make this based on master, but then we had to fix some conflicts later


UziTech (changes requested):
rebase after #160 is merged and add tests

aminya:
I opened an issue for the additional test #164. Feel free to take it. I will merge this PR as the change itself is trivial.

PR #162, opened 2021-06-12
feat: apply additionalTextEdits after completion

aminya:
See https://microsoft.github.io/language-server-protocol/specification#textDocument_completion

```ts
	/**
	 * An optional array of additional text edits that are applied when
	 * selecting this completion. Edits must not overlap (including the same
	 * insert position) with the main edit nor with themselves.
	 *
	 * Additional text edits should be used to change text unrelated to the
	 * current cursor position (for example adding an import statement at the
	 * top of the file if the completion item will insert an unqualified type).
	 */
```

### DIFF
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -35,7 +35,6 @@ import {
 } from "./server-manager.js"
 import { Disposable, CompositeDisposable, Point, Range, TextEditor } from "atom"
 import * as ac from "atom/autocomplete-plus"
-import { SuggestionBase } from "./types/autocomplete-extended"
 import { basename } from "path"
 
 export { ActiveServer, LanguageClientConnection, LanguageServerProcess }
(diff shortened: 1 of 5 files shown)


UziTech (changes requested):
just needs tests

lib/auto-languageclient.ts
```diff
@@ -638,7 +638,10 @@ export default class AutoLanguageClient {
       excludeLowerPriority: false,
       filterSuggestions: true,
       getSuggestions: this.getSuggestions.bind(this),
-      onDidInsertSuggestion: this.onDidInsertSuggestion.bind(this),
+      onDidInsertSuggestion: (event) => {
+        AutocompleteAdapter.applyAdditionalTextEdits(event)
```
UziTech, line 642: Why is this called here instead of in the `onDidInsertSuggestion` function?

aminya: Because this is part of the LSP. The `onDidInsertSuggestion` is independent of the LSP. It is a method that people can override for custom behavior.

UziTech (approved):

PR #183, opened 2021-08-05
fix(convert/pathToUri): Improve the encoding of non-alphanumeric paths other than the file protocol

ayame113:
resolve ayame113/atom-ide-deno#74

There is a difference in the conversion process between `Convert.uriToPath` and `Convert.pathToUri`.

|function|(arguments -> retuen value)|
|---|---|
|`uriToPath()`|`https://foo.bar/%40`->`https://foo.bar/%40`<br>`https://foo.bar/%2540`->`https://foo.bar/%2540`|
|`pathToUri()`|`https://foo.bar/%40`->`https://foo.bar/%2540`|

The decoding process of urls other than file: was changed by https://github.com/atom-community/atom-languageclient/commit/6f1e149c052888ff00c26b79be66e7bc7109d84f, but it seems that the encoding process was not changed.
This PR modifies the encoding process for paths other than file: .

### DIFF
--- a/test/convert.test.ts
+++ b/test/convert.test.ts
@@ -50,6 +50,7 @@ describe("Convert", () => {
     })
 
     it("does not encode Windows drive specifiers", () => {
+      setProcessPlatform("win32")
       expect(Convert.pathToUri("d:\\ee\\ff.txt")).toBe("file:///d:/ee/ff.txt")
     })
 
--- a/test/convert.test.ts
+++ b/test/convert.test.ts
@@ -52,9 +52,10 @@ describe("Convert", () => {
     it("does not encode Windows drive specifiers", () => {
       // This test only succeeds on windows. (Because of the difference in the processing method of drive characters)
       // However, it is enough to test the windows drive character only on windows.
-      if (process.platform === "win32") {
-        expect(Convert.pathToUri("d:\\ee\\ff.txt")).toBe("file:///d:/ee/ff.txt")
+      if (process.platform !== "win32") {
+        pending("Only test on windows")
       }
+      expect(Convert.pathToUri("d:\\ee\\ff.txt")).toBe("file:///d:/ee/ff.txt")
     })
 
     it("URI encodes special characters", () => {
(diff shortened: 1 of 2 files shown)


ayame113:
Oops, the test failed, so I'll fix it.

test/convert.test.ts
```diff
@@ -38,7 +50,11 @@ describe("Convert", () => {
     })
 
     it("does not encode Windows drive specifiers", () => {
-      expect(Convert.pathToUri("d:\\ee\\ff.txt")).toBe("file:///d:/ee/ff.txt")
+      // This test only succeeds on windows. (Because of the difference in the processing method of drive characters)
+      // However, it is enough to test the windows drive character only on windows.
+      if (process.platform === "win32") {
+        expect(Convert.pathToUri("d:\\ee\\ff.txt")).toBe("file:///d:/ee/ff.txt")
+      }
     })
```
ayame113, line 58: `new URL("d:\\ee\\ff.txt", "file://")` seems to return different results for windows and others. I changed this because I thought it would be enough if the test on the drive character was successful only on windows.

test/convert.test.ts
```diff
@@ -38,7 +50,11 @@ describe("Convert", () => {
...
 
     it("does not encode Windows drive specifiers", () => {
-      expect(Convert.pathToUri("d:\\ee\\ff.txt")).toBe("file:///d:/ee/ff.txt")
+      // This test only succeeds on windows. (Because of the difference in the processing method of drive characters)
+      // However, it is enough to test the windows drive character only on windows.
+      if (process.platform === "win32") {
+        expect(Convert.pathToUri("d:\\ee\\ff.txt")).toBe("file:///d:/ee/ff.txt")
+      }
```
UziTech, line 57:
```suggestion
      // This test only succeeds on windows. (Because of the difference in the processing method of drive characters)
      // However, it is enough to test the windows drive character only on windows.
      if (process.platform === "win32") {
        pending("Only test on windows")
      }
      expect(Convert.pathToUri("d:\\ee\\ff.txt")).toBe("file:///d:/ee/ff.txt")
```

ayame113: Thank you for your review.
Fixed to use pending ().

However, the test is failing on unrelated changes.

UziTech (approved):

aminya (approved):

PR #207, opened 2022-07-21
Use textEdit of completionItem unconditionally

machitgarha:
### Identify the Bug

[php-ide-serenata#527](https://github.com/Gert-dev/php-ide-serenata/pull/527#issuecomment-1187717251)

### Description of the Change

The description is explained in the commit. Please read that beforehand.

Give priority to the server's `CompletionItem.textEdit` rather than the `replacementPrefix` extracted in the client. These cases shouldn't be different in most cases, but not some. For example, in the case of a trigger character, the server may decide to replace it along with the prefix, but the client ignores it, resulting in wrong output, because `newText` is created on the server and it is the server who actually knows where it must be placed. 

### Alternate Designs

There are many alternatives actually. But the current change is minimal and it only affects if the server set the `CompletionItem.textEdit`.

### Possible Drawbacks

Possibly zero. If a package using this library get impacted by this, then the LSP server it uses is doing something wrong, so a hidden bug will be exposed. But this is a rare case, at least theoretically.

### Verification Process

Tested with `php-ide-serenata`. The mentioned issue is fixed there (e.g. autocompletion of `$this` leads to `$$this`).

### Release Notes

- Fix `CompletionItem.textEdit` of the server being ignored.

### DIFF
--- a/lib/adapters/autocomplete-adapter.ts
+++ b/lib/adapters/autocomplete-adapter.ts
@@ -191,15 +191,12 @@ export default class AutocompleteAdapter {
         )
     )
 
-    // spec guarantees all edits are on the same line, so we only need to check the columns
-    const triggerColumns: [number, number] = [triggerPoint.column, request.bufferPosition.column]
-
     // Setup the cache for subsequent filtered results
     const isComplete = completions === null || Array.isArray(completions) || !completions.isIncomplete
     const suggestionMap = this.completionItemsToSuggestions(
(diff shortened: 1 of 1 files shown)


aminya (changes requested):
The tests fail

machitgarha:
Yes, I realized it. As soon as I find time, I'll look into it. Thanks for your attention.

